PR #4, opened 2020-04-25
Refactored CI

mvegter:


docker-compose.test.yml
```diff
@@ -0,0 +1,11 @@
+version: "3.7"
+
+services:
```
noxi333, line 3: should this not also need a comment? just like `docker-compose.dev.yml`. for sake of consistency.

PR #5, opened 2020-04-25
Feature graceful shutdown

mvegter:


lib/application/interfaces/Server.js
```diff
@@ -41,6 +41,25 @@ class Server {
     async listen() {
         return Promise.reject('The method or operation is not implemented.');
     }
+
+    /**
+     * Returns wether or not to accept any new incoming connections.
+     *
+     * @returns {Boolean} Wether or not to accept any new incoming connections.
+     */
+    shouldAcceptIncomingConnections() {
```
noxi333, line 50: sounds like a setter, without docs it is unclear, maybe:
```suggestion
    IsAcceptingIncomingConnections() {
```

PR #7, opened 2020-04-27
refactor: architecture

mvegter:


lib/database/index.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * Performs connection to the database.
+     *
+     * @returns {Promise} Promise object represents the outcome.
+     */
+    async connect() {
+        // This is a no-op method
+    }
```
noxi333, line 34: Are empty bodies needed? this looks cleaner?
```suggestion
    async connect(); // This is a no-op method
```

mvegter: I would go for the current one with brackets as when the implementation changes it will only show changes in the body of the method instead of the whole function.

lib/application/index.js
```diff
@@ -11,15 +11,69 @@
...
-        repositories,
-    },
-};
-
-module.exports = { ...interfaces, ...implementations };
+const { Application } = require('./interfaces');
+const Database = require('../database');
+const Server = require('../server');
+
+const SHUTDOWN_DELAY_MS = 5 * 1000;
```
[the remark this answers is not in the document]
mvegter: Made it a bit more verbose.

PR #13, opened 2020-04-29
feat: added Adapter

mvegter:


lib/server/controllers/logs.controller.js
```diff
@@ -25,14 +25,32 @@ const { repositories: { LogRepository } } = require('../../database');
  *                        next middleware function.
  * @returns {undefined}
```
noxi333, line 26: does it?

mvegter: Are you referring to the following:
```diff
-      return response.status(400).json({
+      response.status(400).json({
            errors,
        });
+     return;
```

noxi333: line 42 to 44,  should these be documented?

PR #52, opened 2020-05-17
feat: Added possibility to show more or less filters

MauritsioRK:
https://alice.its.cern.ch/jira/browse/O2B-267

### DIFF
--- a/lib/public/components/Filters/index.js
+++ b/lib/public/components/Filters/index.js
@@ -40,7 +40,7 @@ const checkboxFilter = (model, tags) => {
     if (checkboxes.length <= FILTERS_LIMIT) {
         return checkboxes;
     } else {
-        const showMoreFilters = model.logs.getShouldShowMoreFilters();
+        const showMoreFilters = model.logs.shouldShowMoreFilters();
         const toggleFilters = h('button.btn.btn-primary.mv1#toggleMoreFilters', {
             onclick: () => model.logs.toggleMoreFilters(),
         }, ...showMoreFilters ? [iconMinus(), ' Less filters'] : [iconPlus(), ' More filters']);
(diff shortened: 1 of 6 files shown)


mvegter:
@MauritsioRK the following is not working propertly:

- Press 'more filters'
- Select 'maintenance'
- Press 'less filters'
- Press 'more filters'
- Expected maintenance to still be selected

lib/public/components/Filters/index.js
```diff
@@ -38,16 +39,18 @@ const checkboxFilter = (model, tags) => {
...
-            }, 'Meer opties'),
-            ...checkboxes.slice(FILTERS_LIMITS),
-        ])
-        : h('.form-group', checkboxes);
+    if (checkboxes.length <= FILTERS_LIMIT) {
+        return checkboxes;
+    } else {
+        const toggleFilters = h('button.btn.btn-primary.mv1#toggleMoreFilters', {
+            onclick: () => model.logs.toggleMoreFilters(),
+        }, ...model.logs.moreFilters ? [iconMinus(), ' Less filters'] : [iconPlus(), ' More filters']);
```
john-ghatas, line 47: Define a getter for the moreFilters var in the model

john-ghatas:
Also rebase with master to get the latest changes :)

mvegter:
I'm going to either disagree with the change requested from @john-ghatas or the implementation of @MauritsioRK . The naming convention for a boolean getter or function returning a boolean should be prefixed with `is` (with the alternatives of `has`, `can` and `should`).

MauritsioRK:
@mvegter I do agree with @john-ghatas that we should use getter methods for this. I have changed the relevant method names accordingly.

PR #59, opened 2020-05-20
refactor: fully implementend log tree model

mvegter:


lib/public/Model.js
```diff
@@ -59,6 +59,8 @@ export default class Model extends Observable {
                 this.logs.fetchAllLogs();
                 break;
             case 'entry':
+                this.logs.data = [];
+                this.logs.filteredData = [];
```
john-ghatas, line 63: Why initiate the arrays in the route switch case?

lib/public/components/Post/index.js
```diff
@@ -12,19 +12,63 @@
...
+    id: {
+        name: 'ID',
+        visible: true,
+    },
+    authorId: {
+        name: 'Source',
+        visible: true,
+    },
+    creationTime: {
+        name: 'Created',
```
john-ghatas, line 25: Please verify that the datatype test (frontend tests) in the public folder still works

lib/public/Model.js
mvegter: Some old code, has been removed.

lib/public/components/Post/index.js
mvegter: The test for the datatype tests still works (as can be seen by GH actions), if that is what you meant.

lib/public/components/Post/index.js
```diff
@@ -12,19 +12,63 @@
  */
 import { h } from '/js/src/index.js';
 
+const ACTIVE_FIELDS = {
```
john-ghatas, line 15: Any plans to unify this constant and save it in the ActiveColumns/index.js or any other location? Or does this variable server another purpose

john-ghatas:
When resizing to a narrower screen the text overflows in the other postbox with the text contents like so:
image

lib/public/components/Post/index.js
mvegter: I would say that each screen can have it own user preferences, so no this should not be merged.

mvegter:
> When resizing to a narrower screen the text overflows in the other portion of the postbox with the text contents like so:
> image

This applies to all our screens, unless it is an issue on a 'normal' desktop resolution it is not a priority,

PR #65, opened 2020-05-24
Log overview pagination

MauritsioRK:
https://alice.its.cern.ch/jira/browse/O2B-250

TODO:
- Make sure tag filters aren't reset by the pagination (another ticket?)

### DIFF
--- a/lib/public/app.css
+++ b/lib/public/app.css
@@ -1,5 +1,3 @@
-.btn-right { float: right; }
-
 .cell-xs { width: 2rem; }
 .cell-s { width: 4rem; }
 .cell-m { width: 8rem; }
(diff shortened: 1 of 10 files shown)


mvegter:
@MauritsioRK please rebase on master, a fix for the number of logs has been committed.

john-ghatas:
When applying a filter on a page. the selected filter stays selected when going to another page and going back to the page where the filter was applied

image

image

mvegter:
@john-ghatas (, @MauritsioRK) as the tag filtering didn't function as expected and knowing that fixing it would limit/slow down the pagination feature it was decided to not include it in the current pull request.

That does mean that right now the priority should be to properly implement it, see [O2B-285](https://alice.its.cern.ch/jira/browse/O2B-285) for more details.

PR #94, opened 2020-06-04
RemoteData support for the Logs model

MauritsioRK:
https://alice.its.cern.ch/jira/browse/O2B-303

### DIFF
--- a/lib/public/views/Logs/Details/index.js
+++ b/lib/public/views/Logs/Details/index.js
@@ -20,10 +20,9 @@ import PostBox from '../../../components/Post/index.js';
  * @return {vnode} Return the view of the table with the filtering options
  */
 const logDetailScreen = (model) => {
-    const data = model.logs.getData();
-    const error = model.logs.hasErrorOccurred();
+    const data = model.logs.getLogs();
 
-    if (data && data.length !== 0) {
+    if (data.isSuccess()) {
         const id = parseInt(model.router.params.id, 10);
 
         // eslint-disable-next-line require-jsdoc
@@ -48,15 +47,15 @@ const logDetailScreen = (model) => {
             return items;
         };
 
-        return h('', h('.w-100.flex-column', tree(data[0])));
-    } else if (error) {
+        return h('', h('.w-100.flex-column', tree(data.payload[0])));
+    } else if (data.isFailure()) {
         return h('', [
-            h('.danger', 'This log could not be found.'),
-            h('button.btn.btn-primary.mv3', {
+            h('.alert.alert-danger', 'This log could not be found.'),
+            h('button.btn.btn-primary.mv2', {
                 onclick: () => model.router.go('?page=home'),
             }, 'Return to Overview'),
         ]);
-    } else {
+    } else if (data.isLoading()) {
         return spinner();
     }
 };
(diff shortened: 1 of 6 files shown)


lib/public/views/Logs/Details/index.js
```diff
@@ -48,15 +47,18 @@ const logDetailScreen = (model) => {
...
-    } else if (error) {
         return h('', [
-            h('.danger', 'This log could not be found.'),
-            h('button.btn.btn-primary.mv3', {
+            h('h2.mv2', { onremove: () => model.logs.clearLogs() }, 'Log Tree'),
+            h('.w-100.flex-column', tree(data.payload[0])),
+        ]);
+    } else if (data.isFailure()) {
+        return h('', [
+            h('.alert.alert-danger', 'This log could not be found.'),
```
mvegter, line 56: We are not sure what the errors is, so we should use something like:
```js
data.payload.map((error) => h('.alert.alert-danger', h('b', `${error.title}: `), error.detail))
```

lib/public/views/Logs/Logs.js
```diff
@@ -24,18 +24,18 @@ export default class Overview extends Observable {
     constructor(model) {
         super();
         this.model = model;
+
         this.filterCriteria = [];
-        this.data = [];
-        this.filteredData = [];
+        this.logs = RemoteData.NotAsked();
```
mvegter, line 29: You can replace this with the `clearLogs` method

lib/public/views/Logs/Logs.js
```diff
@@ -82,91 +110,24 @@ export default class Overview extends Observable {
...
-            const logId = result.data.id;
+            this.createdLog = RemoteData.success(result.data);
             this.title = '';
             this.text = '';
-            this.error = false;
 
-            await this.fetchOneLog(logId);
-            await this.model.router.go(`/?page=entry&id=${logId}`);
+            await this.model.router.go(`/?page=entry&id=${this.createdLog.payload.id}`);
             this.notify();
```
mvegter, line 118: Is this notify still needed if we are going to redirect?

lib/public/views/Logs/Logs.js
```diff
@@ -334,4 +315,33 @@ export default class Overview extends Observable {
...
+    setText(text) {
+        this.text = text;
+        this.notify();
+    }
+
+    /**
+     * Sets all data related to the Logs to `NotAsked`.
+     * @returns {undefined}
+     */
+    clearLogs() {
```
mvegter, line 343: If the page is destroyed, shouldn't we also reset the paginiation options here?

lib/public/views/Logs/Details/index.js
MauritsioRK: Done

lib/public/views/Logs/Logs.js
MauritsioRK: Done

lib/public/views/Logs/Logs.js
MauritsioRK: I had a look, and it did not seem detrimental - removed

lib/public/views/Logs/Logs.js
MauritsioRK: Good point - done

PR #95, opened 2020-06-04
Added Markdown preview in the box

john-ghatas:
https://alice.its.cern.ch/jira/secure/RapidBoard.jspa?rapidView=344&projectKey=O2B&view=detail&selectedIssue=O2B-248
Note: Adding files is not yet supported because of the same-origin policy


lib/public/index.html
```diff
@@ -11,10 +11,64 @@
...
     <link rel="stylesheet" href="/app.css">
-
+    
     <script src="/js/src/compatibility.js"></script>
 </head>
 
+  <!-- ----------------------------------------------------------------------------------- -->
+  <!-- CodeMirror -->
+
+  <link rel="stylesheet" href="./assets/SmartEditor/codemirror/lib/codemirror.css">
```
mvegter, line 21: Do not use relative paths here

lib/public/index.html
```diff
@@ -34,11 +88,11 @@
...
-        const model = new Model();
+        const model = new Model(HyperMD);
         const debug = true; // shows when redraw is done
         mount(document.body, view, model, debug);
 
         // Expose model to interact with it the browser's console
         window.model = model;
     </script>
-</body>
+</body>
```
mvegter, line 98: restore the empty line

lib/public/index.html
```diff
@@ -11,10 +11,64 @@
     <link rel="shortcut icon" href="/favicon.ico" />
     <link rel="stylesheet" href="/css/src/bootstrap.css">
     <link rel="stylesheet" href="/app.css">
-
+
```
mvegter, line 14: cleanup here too, please

lib/public/views/Logs/Create/index.js
```diff
@@ -25,6 +25,13 @@ const createScreen = (model) => {
     const errorMessages = model.logs.getErrorMessages()
         || [{ detail: 'There was a problem with posting the log, try again later' }];
 
+    const preview = h('', [
```
mvegter, line 28: Shouldn't this be an arrow function instead of a variable?

lib/public/Model.js
```diff
@@ -58,6 +61,8 @@ export default class Model extends Observable {
      * @returns {vnode} The page to be loaded
      */
     handleLocationChange() {
+        // Clear the editor before navigating
+        this.logs.clearEditor();
```
mvegter, line 65: Shouldn't this be in a onremove method?

lib/public/index.html
john-ghatas: Any specific reason for that?

lib/public/views/Logs/Create/index.js
john-ghatas: Done

lib/public/index.html
john-ghatas: Done

lib/public/Model.js
john-ghatas: Calling the onremove like this: 
image
Doesn't clear the editor (it's never called upon

john-ghatas:
It's a mix with minified js files or full fledged js files (depending on the plug), couldn't find any minified versions so far.

lib/public/index.html
mvegter: As using relative paths can break while absolute paths wont.

lib/public/Model.js
mvegter: Then it is most likely on the wrong component as the onremove is called in the tag overview/detail and in logs (see the open PR).

mvegter:
I found minified version by changing the .js extension to .min.js of the imports in https://github.com/laobubu/HyperMD/blob/master/docs/examples/ai1.html

lib/public/Model.js
john-ghatas: Fixed :)

lib/public/index.html
john-ghatas: Done as well

john-ghatas:
Added the minified versions as much as I could, see [the latest commit](https://github.com/AliceO2Group/Bookkeeping/pull/95/commits/41dfb5db086ca30006518d4254a2472122bede81)

.nycrc
```diff
@@ -7,6 +7,7 @@
         "lib/main.js",
         "lib/database/migrations/*.js",
         "lib/database/seeders/*.js",
-        "test/**/*.js"
+        "test/**/*.js",
+        "lib/public/assets/**"
```
mvegter, line 11: Sort these in alphabetical order

john-ghatas: Done

lib/public/Model.js
john-ghatas: @mvegter

PR #113, opened 2020-06-09
fix: Set width to auto in the markdown box when creating an entry

john-ghatas:
Instructions to verify: Go to the create screen and create a title like: # Sample Text, hit enter and the grey line should cover the whole width of the box now

### DIFF
--- a/lib/public/views/Logs/Create/index.js
+++ b/lib/public/views/Logs/Create/index.js
@@ -63,7 +63,7 @@ const createScreen = (model) => {
             }, title),
         ]),
         h('h3.black.line-break: auto', 'Text'),
-        h('.shadow-level2.w-90', mdBox(model, text)),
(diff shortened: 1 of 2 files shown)


john-ghatas:
The GitHub machines can't build the test suite for the coverage at the moment...

lib/public/views/Logs/Create/index.js
```diff
@@ -63,7 +63,7 @@ const createScreen = (model) => {
             }, title),
         ]),
         h('h3.black.line-break: auto', 'Text'),
-        h('.shadow-level2', mdBox(model, text)),
+        h('.shadow-level2.w-90', mdBox(model, text)),
```
mvegter, line 66: Why not `w-100`? At moment 10% is lost on the right of the screen.

PR #118, opened 2020-06-11
Restore filtering by tags

MauritsioRK:
https://alice.its.cern.ch/jira/browse/O2B-285

### DIFF
--- a/lib/public/app.css
+++ b/lib/public/app.css
@@ -65,6 +65,10 @@ html, body {
     color: #491217;
 }
 
+.form-group-header {
+    margin-bottom: 0.25rem;
+}
+
 .nav {
     display: -ms-flexbox;
     display: flex;
--- a/lib/public/components/Filters/index.js
+++ b/lib/public/components/Filters/index.js
@@ -45,7 +45,6 @@ const filterOperationRadioButtons = (model) => h('.form-group-header.flex-row', 
 const tagCheckboxes = (model, tags) => {
     const checkboxes = tags.map((tag, index) => {
         const isChecked = model.logs.isTagInFilter(tag.id);
-        const count = 1; // TODO: Restore the dynamic count for logs related to this tag
         return h('.form-check', [
             h('input.form-check-input', {
                 onclick: () => isChecked ? model.logs.removeFilter(tag.id) : model.logs.addFilter(tag.id),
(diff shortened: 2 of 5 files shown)


MauritsioRK:
@mvegter 
>     * switching between table pages causes the tag filter to be re-rendered for which is quite annoying (due to the flickering).
Should the loading spinner only appear during the first load, or do you just mean to remove the `isSuccess` check from the table render?

>     * Seems like filtering does not work completely, notice the missing tags
Do you mean the display of the "Tags" column? That is a backend issue, I think - the tags shown in the result are all I'm getting from the tags array of a log when applying a [tag][values] filter. Unless I'm using this incorrectly.
See http://localhost:4000/api/logs?filter[tag][values]=1&filter[tag][operation]=or

mvegter:
The fetching of table data does not affect the filters as far as I know. So they should have independent loaders (, one could say that the initial load of both is a single loader).

Based on the simplicity of the reproduction scenario I would have expected you to have caught this earlier and created a fix in this (or an earlier) pull request. A bug affecting the story you work on, be it front-end/back-end/infrastructure, should be reported and fixed first as it affects your story.

MauritsioRK:
> The fetching of table data does not affect the filters as far as I know. So they should have independent loaders (, one could say that the initial load of both is a single loader).

I'm not sure what you mean. The filtering **is** dependent on the filters, the tag filtering is included in the API call. They can't be independent.

> Based on the simplicity of the reproduction scenario I would have expected you to have caught this earlier and created a fix in this (or an earlier) pull request. A bug affecting the story you work on, be it front-end/back-end/infrastructure, should be reported and fixed first as it affects your story.

The case of this ticket as I see it is solely the front-end functionality of filtering based on the back-end call that was already created. I also have limited knowledge of the inner workings of the backend, so I fear it'd take me hours to figure out the problem.

mvegter:
> I'm not sure what you mean. The table data **is** dependent on the filters, the tag filtering is included in the API call. They can't be independent.

You are right, table data is dependent on the filters. Just not the other way around as I said.

> The case of this ticket as I see it is solely the front-end functionality of filtering based on the back-end call that was already created. I also have limited knowledge of the inner workings of the backend, so I fear it'd take me hours to figure out the problem.

This is called software engineering.

PR #128, opened 2020-06-18
Subsystem overview

DdeHoog:


test/public/subsystems/detail.test.js
```diff
@@ -0,0 +1,135 @@
...
+
+        pti.write([...jsCoverage, ...cssCoverage].filter(({ url = '' } = {}) => url.match(/\.(js|css)$/)), {
+            includeHostname: false,
+            storagePath: './.nyc_output/lib/public',
+        });
+        await browser.close();
+    });
+
+    it('subsystem detail loads correctly', async () => {
+        await page.goto(`${url}/?page=subsystems&id=1`);
```
mvegter, line 70:
```suggestion
        await page.goto(`${url}/?page=subsystem&id=1`);
```

lib/public/views/Subsystems/Details/index.js
```diff
@@ -0,0 +1,84 @@
...
+                        onclick: () => model.router.go(`?page=entry&id=${entry.id}`),
+                    })),
+                    Failure: (payload) => payload.map((error) => h('.alert.alert-danger', error.title)),
+                }),
+            },
+        };
+
+        return [
+            h('h2.mv2', { onremove: () => model.subsystems.clearSubsystem() }, `Subsystem: ${data.payload.name}`),
+            h('.w-100', [
```
mvegter, line 66: Isn't this exactly the same as the one in Tags? Perhaps a generic component can be created.

PR #129, opened 2020-06-18
Fix for missing tags when filtering

MauritsioRK:
https://alice.its.cern.ch/jira/browse/O2B-318

I know this one took a while... it was a lot of research to be done, so apologies for the wait.


test/usecases/log/GetAllLogsUseCase.test.js
```diff
@@ -43,6 +43,19 @@ module.exports = () => {
         }
     });
 
+    it('should return logs with a full tag collection regardless of filter', async () => {
```
MauritsioRK, line 46: note to self: This test should probably work the other way around (filtered result first, then compare against unfiltered)

mvegter:
So this "might" work as there are no tests added/changed to proof it didn't work and now does. Luckily I already created some so please include them.

LogTag should be LogTags, to match the database table name.

MauritsioRK:
> 
> 
> So this "might" work as there are no tests added/changed to proof it didn't work and now does. Luckily I already created some so please include them.
> 
> LogTag should be LogTags, to match the database table name.

Doesn't GetLogUseCase test cover this?

mvegter:
You are missing E2E tests, branch is test/report-tags

lib/database/models/index.js
```diff
@@ -14,6 +14,7 @@
 const EpnRoleSession = require('./epnrolesession');
 const FlpRole = require('./flprole');
 const Log = require('./log');
+const LogTags = require('./logtag');
```
mvegter, line 17:
```suggestion
const LogTags = require('./logtags');
```

lib/usecases/log/GetAllLogsUseCase.js
```diff
@@ -54,18 +52,26 @@ class GetAllLogsUseCase {
...
+                subQueryBuilder.where('tagId').oneOf(...tag.values);
+                subQueryBuilder.orderBy('logId', 'asc');
+                let logTags;
+
                 switch (tag.operation) {
                     case 'and':
-                        queryBuilder
-                            .setModel(Log)
-                            .whereAssociation('tags', 'id').allOf(...tag.values);
+                        logTags = await await TransactionHelper
```
mvegter, line 62: These should be include the in the other transaction helper.

lib/usecases/log/GetAllLogsUseCase.js
```diff
@@ -54,18 +52,26 @@ class GetAllLogsUseCase {
             }
 
             if (tag) {
+                const subQueryBuilder = new QueryBuilder();
+                subQueryBuilder.where('tagId').oneOf(...tag.values);
+                subQueryBuilder.orderBy('logId', 'asc');
```
mvegter, line 57: You can actually chain these methods

lib/usecases/log/GetAllLogsUseCase.js
MauritsioRK: What do you mean exactly? Include what in which other transaction helper?

mvegter: Included in the transaction helper on line 89

MauritsioRK: Ahh okay, gotcha

lib/database/seeders/20200511151016-log-tags.js
```diff
@@ -37,6 +37,22 @@ module.exports = {
             log_id: 3,
             tag_id: 6,
         },
+        {
+            log_id: 3,
```
mvegter, line 41: This is a duplicate, thus constraint

lib/database/seeders/20200511151016-log-tags.js
```diff
@@ -37,6 +37,22 @@ module.exports = {
...
+            tag_id: 1,
+        },
+        {
+            log_id: 2,
```
mvegter, line 45: Same here

lib/database/models/logtags.js
```diff
@@ -0,0 +1,29 @@
...
+    const LogTags = sequelize.define('LogTags', {
+        logId: {
+            type: Sequelize.INTEGER,
+        },
+        tagId: {
+            type: Sequelize.INTEGER,
+        },
+    });
+
+    LogTags.removeAttribute('id');
```
mvegter, line 26: I think this is not required if both logId and tagId have the `primary: true` flag

MauritsioRK: Unfortunately it is required, otherwise I get the error `Unknown column 'id' in 'field list'` when requesting from this model. I do admit it's kind of a patchy fix, but that's all I could find

mvegter:
GetLogsByTagUseCase this one should be adapted to use the GetAllLogsUseCase somehow that will fix one of the failing tests

MauritsioRK:
For some reason my local tests aren't running correctly, so finding the problems is a little hard right now. Need to look into this for some longer time.

lib/usecases/tag/GetLogsByTagUseCase.js
```diff
@@ -33,15 +30,11 @@ class GetLogsByTagUseCase {
...
      */
     async execute(dto) {
-        const queryBuilder = new QueryBuilder();
         const { params } = dto;
         const { tagId } = params;
 
-        queryBuilder.where('id').is(tagId);
-        return TransactionHelper.provide(async () => {
-            const tag = await TagRepository.findOne(queryBuilder);
-            return tag ? LogRepository.findAllByTagId(tag.id) : null;
```
mvegter, line 43: @MauritsioRK , The findOne step is required to determine if a log actually exists.

lib/usecases/tag/GetLogsByTagUseCase.js
```diff
@@ -38,10 +38,14 @@ class GetLogsByTagUseCase {
         const { tagId } = params;
 
         queryBuilder.where('id').is(tagId);
-        return TransactionHelper.provide(async () => {
-            const tag = await TagRepository.findOne(queryBuilder);
-            return tag ? LogRepository.findAllByTagId(tag.id) : null;
```
mvegter, line 43: The `findAllByTagId` method of LogRepository can be removed.

PR #130, opened 2020-06-19
Added collapse feature in generic table component

john-ghatas:
https://alice.its.cern.ch/jira/secure/RapidBoard.jspa?rapidView=344&projectKey=O2B&view=detail&selectedIssue=O2B-313
image
image


lib/public/app.css
```diff
@@ -138,4 +138,3 @@ html, body {
 .tab-content > .active {
     display: block;
 }
-
```
john-ghatas, line 141: EOF is there in the local copy

mvegter:
It doesn't work right? Your first image shows that titles that are too long still affect row height.

john-ghatas:
Row height is unaffected now, it stays unaffected even when resizing horizontally

mvegter:
Does it still work with   size: 'cell-xl' and   size: 'cell-xs'?

john-ghatas:
It was set on XL already, XS shows no issues either

john-ghatas:
Title column set on XS
image

mvegter:
How can a XS column be wider than an L column

john-ghatas:
Because the text is limited to one line with CSS, thus filling all the space and taking extra space in the table

mvegter:
Than what is the point of giving a size property if the content decides the width?

john-ghatas:
Removing it has no consequence and makes it auto resize depending on the set treshold :)

mvegter:
So that is a no. The word-breaking should be based on the column width not the column width based on the content or another property. We know the width of the column so the word-breaking can be calculated.

john-ghatas:
Added extra CSS classes for this specific use case S-XXL and cleaned up some code

lib/public/views/Logs/ActiveColumns/index.js
```diff
@@ -15,7 +15,9 @@ const ACTIVE_COLUMNS = {
     title: {
         name: 'Title',
         visible: true,
-        size: 'cell-l',
+        expand: true,
+        expandTreshold: 40,
```
mvegter, line 19: This should either be calculated/dynamic or removed. This creates behavior such as the following (where more could/should have displayed but didn't):
image
image

john-ghatas: I cannot find another way to implement this with CSS/Javascript the treshhold is there to check for the text length to enable the plus/minus

john-ghatas: This is resolved in the latest version :)

destroy-docker.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
mvegter, line 1: Remove this file

lib/public/Model.js
```diff
@@ -29,13 +29,18 @@ export default class Model extends Observable {
...
-    constructor(HyperMD, CompleteEmoji) {
+    constructor(HyperMD, window, document, CompleteEmoji) {
         super();
-        // Bind HyperMD
+        // Bind HyperMD, window and document
         this.HyperMD = HyperMD;
         this.CompleteEmoji = CompleteEmoji;
+        this.document = document;
+        this.window = window;
+        this.CompleteEmoji = CompleteEmoji;
```
mvegter, line 43: Remove this

lib/public/components/Table/content.js
```diff
@@ -10,31 +10,73 @@
  * granted to it by virtue of its status as an Intergovernmental Organization
  * or submit itself to any jurisdiction.
  */
-
-import { h } from '/js/src/index.js';
+import { h, iconPlus, iconMinus } from '/js/src/index.js';
```
mvegter, line 13: Restore empty line

lib/public/components/Table/content.js
```diff
@@ -10,31 +10,73 @@
...
+    const isExpanded = model.logs.isColumnExpanded(rowId, value.name);
+    const base = `.${value.size}#${columnId}`;
+
+    return h(`td${base}`, {
+        onupdate: () => {
+            if (value.expand) {
+                const element = model.document.getElementById(`${columnId}-text`);
+                const minimalHeight = model.logs.getMinimalColumnHeight(rowId, value.name);
+                if (
+                    !canExpand && (element.scrollHeight > element.offsetHeight || element.scrollHeight > minimalHeight)
```
mvegter, line 37: Store the (sub-)expression in a variable instead of this weird format

lib/public/components/Table/index.js
```diff
@@ -34,12 +34,13 @@ const filterKeysByVisibility = (keys) => {
  * @param {Object} keys The full collection of API keys and their corresponding header values
  * @param {Function} params Additional element parameters, wrapped in a function
  * @returns {vnode} Return the total view of the table to rendered
+ * @param {Object} model The global model object
```
mvegter, line 37: Order

lib/public/components/common/markdown.js
```diff
@@ -20,10 +20,10 @@
  * @param {Object} size The size of the box, in case the standard is too little or too big
  * @returns {vnode} The converted editor
  */
-const setMarkDownBox = (textAreaId, model, changeHandler, readOnly, size) => {
+const setMarkDownBox = (textAreaId, model,
```
mvegter, line 23: Not sure if this change is needed

lib/public/views/Logs/Logs.js
```diff
@@ -115,7 +120,6 @@ export default class Overview extends Observable {
         const response = await fetchClient('/api/logs', {
             method: 'POST',
             headers: {
-                Accept: 'application/json',
```
mvegter, line 118: Any reason for this?

lib/public/views/Logs/Logs.js
```diff
@@ -396,4 +403,82 @@ export default class Overview extends Observable {
...
+    /**
+     * Add eligble columns for collapse to the array in the model
+     * @param {String} rowId The rowId being collapsed
+     * @param {String} name The name of the column to be collapsed
+     * @param {Integer} height Minimal height of the column
+     * @returns {undefined}
+     */
+    addCollapsableColumn(rowId, name, height) {
+        const existingColumn = this.collapsableColumns
+            .find((element) => element.rowId === rowId && element.name === name);
```
mvegter, line 416: Seeing this logic `element.rowId === rowId && element.name === name` multiple times, can't this be in any cleaner form.

lib/public/views/Logs/Logs.js
```diff
@@ -35,6 +35,11 @@ export default class Overview extends Observable {
 
         this.isPreviewActive = false;
         this.editors = [];
+
+        this.collapsableColumns = [];
+        this.collapsedColumns = [];
+
+        this.model.window.addEventListener('resize', () => this.notify());
```
mvegter, line 42: Wouldn't it be better to have this in the `index.html` as table resizing and stuff is on multiple screens?

lib/public/views/Logs/ActiveColumns/index.js
```diff
@@ -15,7 +15,8 @@ const ACTIVE_COLUMNS = {
     title: {
         name: 'Title',
         visible: true,
-        size: 'cell-l',
+        expand: true,
```
mvegter, line 18: Restore the original value.

lib/public/views/Logs/Logs.js
john-ghatas: The most central place where we can put this code so it applies to all pages is the main Model itself, guess that would work as well?

lib/public/views/Logs/Logs.js
MauritsioRK: @john-ghatas I think this was caused by rebase at some point

lib/public/components/Table/content.js
john-ghatas: Done

lib/public/components/common/markdown.js
john-ghatas: Done

lib/public/views/Logs/Logs.js
john-ghatas: Done as well

lib/public/views/Logs/ActiveColumns/index.js
john-ghatas: Done as well (shows up correct locally)

lib/public/components/common/markdown.js
```diff
@@ -22,8 +22,7 @@
  */
 const setMarkDownBox = (textAreaId, model, changeHandler, readOnly, size) => {
     const { HyperMD, CompleteEmoji } = model;
-    const { location = '', name = '' } = changeHandler;
-
+    const { location, name } = changeHandler;
```
mvegter, line 25: This is still wrong

PR #164, opened 2020-07-17
Run information

DdeHoog:


lib/server/controllers/runs.controller.js
```diff
@@ -110,21 +92,33 @@ const patchLog = (request, response, next) => {
...
+    }
 };
 
 module.exports = {
-    index,
-    create,
-    patch,
+    getRunById,
+    listRuns,
     patchLog,
```
MauritsioRK, line 123: I think the patchLog method can also be removed

package.json
```diff
@@ -21,6 +21,7 @@
     "mariadb": "2.4.1",
     "multer": "1.4.2",
     "sequelize": "6.2.4",
+    "sequelize-cli": "6.2.0",
```
MauritsioRK, line 24: I don't think this should be in here. It's needed as a global dependency but not here. Please run `npm uninstall sequelize-cli`

spec/openapi-source.yaml
```diff
@@ -166,6 +166,34 @@ paths:
           $ref: '#/components/responses/NotFound'
       tags:
         - log
+  /runs:
+    get:
+      operationId: listRuns
+      summary: List all runs
+      parameters:
+        - $ref: '#/components/parameters/Pagination'
```
MauritsioRK, line 174: I would also add sorting to these parameters. You'll probably want to make the current "SortLogs" ref name more generic, for example "Sorting".

lib/usecases/run/GetAllRunsUseCase.js
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Executes this use case.
+     *
+     * @param {Object} dto The GetAllRuns DTO which contains all request data.
+     * @returns {Promise} Promise object represents the result of this use case.
+     */
+    async execute(dto = {}) {
+        const queryBuilder = new QueryBuilder();
+        const { query = {} } = dto;
+        const { page = {}, sort = { id: 'desc' } } = query;
```
MauritsioRK, line 38: I think the default value for this should be `runNumber` instead of `id`

test/usecases/run/GetAllRunsUseCase.test.js
```diff
@@ -0,0 +1,69 @@
...
+            .execute(getAllRunsDto);
+        expect(filteredResult.runs.length).to.be.greaterThan(0);
+        const [firstFilteredRun] = filteredResult.runs;
+
+        const unfilteredResult = await new GetAllRunsUseCase()
+            .execute();
+        const firstUnfilteredRun = unfilteredResult.runs.find((run) => run.id === firstFilteredRun.id);
+
+        expect(firstUnfilteredRun.tags).to.deep.equal(firstFilteredRun.tags);
+    });
```
MauritsioRK, line 59: Can be removed

PR #183, opened 2020-08-27
Log overview sorting

MauritsioRK:
https://alice.its.cern.ch/jira/browse/O2B-217


MauritsioRK (commented):
There is still a fairly nasty bug present within this PR related to sorting in combination with pagination. Unfortunatenly due to v1 priorities this will have be to be fixed at a later date.

DdeHoog (approved):

PR #202, opened 2020-09-28
Bug/O2B-343 Filters are connected to both Logs and Runs

Ryukishi:
Created a new filter folders with a new filter for Runs.

The filters of Runs and Logs are now seperated from one another.


Ryukishi:
Hi Damian, for this issue I created a new filter specifically for Runs. I made this without any functionality so that the issue is solved. But this leaves a filter for Runs that remains untested since there are no requirements for the filter that I can implement in order to test it. That is why I am not hitting the target for CodeCov.

DdeHoog (approved):

PR #259, opened 2020-12-03
Use CMake targets and transitive deps in C++ binding

awegrzyn:
The relevant alidist recipe: https://github.com/alisw/alidist/pull/2666
As soon as it's merged the compilation will be as simple as: `aliBuild build Bookkeeping`

Still to be fixed:
- [ ] OpenSSL transitive dependency from `cpprestsdk::cpprest` target, temporary workaround in L144


rummens1337:
Tested and doesn't break the manual functionality.
Assuming alidist recipe will be merged soon.

rummens1337 (commented):
Approved

rummens1337 (commented):
Approved

rummens1337 (approved):
Approved. third time's the charm :)

PR #267, opened 2020-12-14
including some timeouts for UI tests

ekertt:


rummens1337 (approved):
Adding timeouts for now, will change timeout to without timeout after merges.

PR #275, opened 2021-01-04
Feature/O2B-373-log-entries-table

Nightfeather404:
- Use tooltip to display full title of a log
- Use tooltip to display long list of tags
- Fix hovering run column changes its width bug


ekertt (commented):
gemerged

PR #281, opened 2021-01-13
O2 b 381

ekertt:


rummens1337 (approved):
Tested locally on Windows & Ubuntu 20.04.
openapi-source.yaml is to be deleted from go-client and update it with the existing one in bookkeeping->spec->openapi-source.yaml.

PR #286, opened 2021-01-14
Use the latest framework to ged rid of sqlite3 dependency

awegrzyn:


awegrzyn:
@TimoWilken @DdeHoog We need this for the FLP Suite release as there was a major issue with 1 of Framework dependencies.

PR #336, opened 2021-04-01
Feature/O2B-304

io112:
Posts detailed and collapsed view on the detailed log overview screen.


DimitrySteenkamer (approved):
The log collapse system works and the tests as well.

PR #349, opened 2021-04-19
fixed doubling log entries

io112:


DimitrySteenkamer (changes requested):
Both the log-overview and tag-overview infinite mode selector now works correctly. When infinite rows per page are selected. It now correctly adds the remaining rows to the already existing rows.

But the run-overview still has the same problem. When infinite rows per page are selected, it adds all the rows to the already existing ones.

DimitrySteenkamer (commented):
It works, the bug has been fixed in the tag, runs and logs overview

DimitrySteenkamer (approved):

io112:
There is some problems going from the current master state after [db-namig-fix was merged](https://github.com/AliceO2Group/Bookkeeping/runs/2426350613).

PR #354, opened 2021-05-05
Add GH workflow to create bundled TARballs on release

awegrzyn:
This is required to install Bookkeeping without connection to npm registry.

After creating a release (tag is not enough) this workflow will automatically upload TARBall to assets with all necessary deps.


DimitrySteenkamer (approved):
Agreed, but the openapi-validator shows an error

PR #371, opened 2021-06-10
[O2B-386] Return arrayResponse, response and err to the caller

DimitrySteenkamer:


graduta:
Done as part of https://github.com/AliceO2Group/Bookkeeping/pull/573/files

PR #374, opened 2021-06-11
Updated aliceo2/bookkeeping version and aliceo2/web-ui

DimitrySteenkamer:


skanibani (approved):
Version update with new version of WebUI which fixes a minor token bug, see changelog

PR #377, opened 2021-06-21
Added the ArrayOfFlpResponse to Example.cpp and comments added.

DimitrySteenkamer:


cpp-api-client/src/Example.cpp
```diff
@@ -47,6 +49,7 @@ int main(int argc, char const *argv[])
     api->flpAdd("flp-2", "localhost", runNumber);
 
     // Update flp
+    // First parameter is the id of the FLP you want to update.
```
vascobarroso, line 52: @DimitrySteenkamer, how does the caller of `flpUpdateCounters` know the "id of the FLP you want to update" ? Is it returned by `flpAdd` ?

DimitrySteenkamer: It's not returned by flpAdd. I've added ArrayOfFlpResponse to inspect all the id's, but the caller probably can't access that. Would it be a solution if flpAdd returns the flpId?

vascobarroso: Maybe @sy-c has a better idea, but I think it makes sense for the method that creates an entry to return the id of the new entry (particularly if it is an auto-generated ID) for future use.

PR #404, opened 2021-08-23
[O2-2512] Make /api/runs route public

awegrzyn:
Until token management is ready we need to make sure that all subsystem will be able to access `/api/runs` route


awegrzyn:
@DimitrySteenkamer We need this until token management is fully working

PR #430, opened 2021-10-27
[O2B-432] Modified Regex, extended timeout on runs filter.

whdegraaf:


graduta (changes requested):
For me the filtering does not work based on run number (before the changes or after the changes are applied). 
Let's review together during the meeting as it seems to be a more complex issue

PR #438, opened 2021-11-18
[O2B-447] Removed main panel and modified router params to point to logs panel.

whdegraaf:
Tested on own desktop within browser and ran local chai tests without issues.

### DIFF
--- a/lib/public/views/Tags/Details/index.js
+++ b/lib/public/views/Tags/Details/index.js
@@ -31,7 +31,7 @@ const tagDetails = (model) => {
     }
 
     if (!model.router.params.panel) {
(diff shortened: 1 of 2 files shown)


graduta (approved):

graduta (changes requested):
As `main` tab will no longer be the default, a test will fail `test/public/tags/detail.test.js`
The scenario in which the test would switch between `main` and `logs` tab needs to be removed  as the main tab does not exist anymore. 
Failing test action can be found here: https://github.com/AliceO2Group/Bookkeeping/runs/4284228324?check_suite_focus=true#step:6:1629

graduta (approved):

PR #440, opened 2021-11-19
[O2B-438] Moved main logs table to the right of the screen. 

4libutt:
…wn by default. Modified tests, but errors are still showing. The test errors are still there but the functionality is correct. I couldn't test it all the way because of certain 500 internal server errors

### DIFF
--- a/lib/public/views/Runs/Details/index.js
+++ b/lib/public/views/Runs/Details/index.js
@@ -82,27 +82,27 @@ const runDetails = (model) => {
         };
 
         return [
-            h('.flex-row', [
-                h('.flex.w-100', [
-                    h('h2.mv2', { onremove: () => model.runs.clearRun() }, `Run #${data.payload.runNumber}`),
+            h('.flex-row.w-100', [
+                h('.flex.w-50', [
+                    h('h2.mv2.mh2', { onremove: () => model.runs.clearRun() }, `Run #${data.payload.runNumber}`),
                     h('.flex-row.mv2', [
-                        h('button.btn.btn-primary.w-15.mh2#create', {
+                        h('button.btn.btn-primary.w-30.mh2#create', {
                             onclick: () => model.router.go(`/?page=log-create&id=${data.payload.runNumber}`),
                         }, 'Add Log Entries'),
-                        h('label.form-check-label.f5.w-11.mh2', 'Add log entries to the run'),
+                        h('label.form-check-label.f5.w-70.mh2', 'Add log entries to the run'),
                     ]),
                     h('.flex-row.mv2', [
-                        h('button.btn.btn-success.w-15.mh2', {
+                        h('button.btn.btn-success.w-30.mh2', {
                             onclick: () => model.runs.updateRunTags(),
                             title: 'This action will overwrite existing tags',
                             disabled: !model.runs.selectedTagsChanged(),
                         }, 'Update Tags'),
-                        h('label.form-check-label.f5.w-15.mh2', { for: 'tags' }, [
+                        h('label.form-check-label.f5.w-70.mh2', { for: 'tags' }, [
                             'Add one or multiple tags to the run. ',
                             '(CTRL + click to select multiple tags).',
                         ]),
                     ]),
-                    h('select#tags-control.form-control.w-30.mh2', {
+                    h('select#tags-control.form-control.w-50.mh2', {
                         multiple: true,
                         onchange: ({ target }) => model.runs.setSelectedTags(target.selectedOptions),
                     }, !tagsAvailable ? h('option', {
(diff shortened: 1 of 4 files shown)


lib/public/views/Runs/Details/index.js
```diff
@@ -86,37 +82,42 @@ const runDetails = (model) => {
...
-                h('button.btn.btn-success.w-15.mh2', {
-                    onclick: () => model.runs.updateRunTags(),
-                    title: 'This action will overwrite existing tags',
-                    disabled: !model.runs.selectedTagsChanged(),
-                }, 'Update Tags'),
-                h('label.form-check-label.f5.w-15.mh2', { for: 'tags' }, [
-                    'Add one or multiple tags to the run. ',
-                    '(CTRL + click to select multiple tags).',
+            h('.flex-row', [
+                h('.flex.w-100', [
```
graduta, line 86: It looks like you are assigning full width to this div and the run details page is placed at the end.
Could you try to split the space in `w-50` for both elements?

Because of this, we will also need to have a look at the inside elements such ash the `select` tag dropdown which currently is set to `w-30` of its parent.

lib/public/views/Runs/Details/index.js
```diff
@@ -86,37 +82,42 @@ const runDetails = (model) => {
...
+                            onclick: () => model.runs.updateRunTags(),
+                            title: 'This action will overwrite existing tags',
+                            disabled: !model.runs.selectedTagsChanged(),
+                        }, 'Update Tags'),
+                        h('label.form-check-label.f5.w-15.mh2', { for: 'tags' }, [
+                            'Add one or multiple tags to the run. ',
+                            '(CTRL + click to select multiple tags).',
+                        ]),
+                    ]),
+                    h('select#tags-control.form-control.w-30.mh2', {
```
graduta, line 105: I believe it would be better if we increase the width to at least 50% of the parent size

lib/public/views/Runs/Details/index.js
```diff
@@ -86,37 +82,42 @@ const runDetails = (model) => {
...
+                        }, 'Add Log Entries'),
+                        h('label.form-check-label.f5.w-11.mh2', 'Add log entries to the run'),
+                    ]),
+                    h('.flex-row.mv2', [
+                        h('button.btn.btn-success.w-15.mh2', {
+                            onclick: () => model.runs.updateRunTags(),
+                            title: 'This action will overwrite existing tags',
+                            disabled: !model.runs.selectedTagsChanged(),
+                        }, 'Update Tags'),
+                        h('label.form-check-label.f5.w-15.mh2', { for: 'tags' }, [
```
graduta, line 100: Moreover, the labels use only `15%` of the space because before they were alone on the entire row. 
I believe they should now be updated together with the buttons. 
Maybe something similar to `w-30` for the buttons to keep the same size and `w-70` for the labels to have the text on one line

graduta (changes requested):

lib/public/views/Runs/Details/index.js
```diff
@@ -86,37 +82,42 @@ const runDetails = (model) => {
...
+                    }, h('a#tagCreateLink', {
+                    }, noTagsText)) :
+                        allTags.isSuccess() ? [
+                            ...allTags.payload.map((tag) => h('option', {
+                                value: tag.id,
+                                selected: selectedTags.includes(tag.id),
+                            }, tag.text)),
+                        ] : h('option', { disabled: true }, 'Loading tags...')),
                 ]),
+                h('.justify-content-end', [runDetail(model, data.payload)]),
```
graduta, line 119: I cannot find this CSS class either in the WebUI or Bookkeeping CSS `justify-content-end`
If you would like to create one with certain properties, feel free to add it to `app.css`

If you would like to add certain CSS styles to this node only you can rewrite it as below:
```
h('.justify-content-end', {
     style: 'display:flex; justify-content: flex-end'
}, [
     runDetail(model, data.payload)
]),
```

graduta (changes requested):

lib/public/views/Runs/Details/index.js
4libutt: Fixed

lib/public/views/Runs/Details/index.js
4libutt: Fixed

graduta (approved):
Good job! 
Feel free to merge at any point and update the ticket

lib/public/views/Runs/Details/index.js
4libutt: Fixed

lib/public/views/Runs/Details/index.js
4libutt: Left it as one node only

PR #441, opened 2021-11-19
[O2B-425] Replies are backend validated

JikHuijberts:
fix(log reply) : Created dto validation that is log/reply specific.

The dto of logs is changed that if there is a parent id, the title will be optional.

Tested valid and invalid inputs with Postman for both log creation and replies.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,6 +3,15 @@ updates:
 - package-ecosystem: npm
   directory: "/"
   schedule:
-    interval: daily
+    interval: weekly
+    day: "tuesday"
+    time: "07:00"
+    timezone: "Europe/Zurich"
   open-pull-requests-limit: 10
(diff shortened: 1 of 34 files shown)


graduta (changes requested):
Looks like the linter job is failing with some new errors in class 'CreateLogDto.js`
Would you be able to have a look at those errors?

Moreover, please have a look at `CreateLogUseCase.test.js` file and try to add a new test for the reply scenario which should check that no title is provided

graduta (changes requested):
One scenario that we did not take into consideration is the fact that now logs will be display with empty string in `Log Page` and `Home Page` (screenshot attached)
image


For now, we would need to add a custom message similar to mails such as `Re: to log <parentLogId>`

In the future and a separate ticket we will look at getting the title of the parent.

graduta (approved):

PR #451, opened 2021-11-24
[O2B-450] CSV run Export not working

Sandervdk:
Fixed error occuring from the csv export by not running the export commands if there is no valid run
Added DB call to collect the runs from the database (it previosly only used the currently visible runs for the export method)
Added visible error on incorrect run number
Added test to verify incorrect run number error

### DIFF
--- a/lib/public/views/Runs/Runs.js
+++ b/lib/public/views/Runs/Runs.js
@@ -196,12 +196,37 @@ export default class Overview extends Observable {
 
     /**
      * Filter runs based on the selected run numbers and fields to prepare it for export
-     * @param {Object} runs The runs data
      * @returns {Object} The Filtered runs data
      */
-    async getFilteredRuns(runs) {
-        const selectedRuns = runs.filter((run) => this.getSelectedRunNumbers().includes(run.runNumber));
-        const filteredRuns = selectedRuns.map((selectedRun) => pick(selectedRun, this.getSelectedRunsFields()));
+    async getFilteredRuns() {
+        this.runs = RemoteData.loading();
+        this.notify();
+
+        const runIds = this.getSelectedRunNumbers().split(',');
+        const foundRuns = [];
+        let response;
+        let result;
+
+        // Loops through every run id that has been filled in and retrieves them from the database
+        for (let i = 0; i < runIds.length; i++) {
+            response = await fetchClient(`/api/runs/${runIds[i]}`, { method: 'GET' });
+            result = await response.json();
+
+            if (!result.data) {
+                this.runs = RemoteData.failure(result.errors || [
+                    {
+                        title: result.error,
+                        detail: result.message,
+                    },
+                ]);
+                this.notify();
+                return '';
+            }
+            foundRuns.push(result.data);
+        }
+
+        this.runs = RemoteData.success();
+        const filteredRuns = foundRuns.map((selectedRun) => pick(selectedRun, this.getSelectedRunsFields()));
         return filteredRuns;
     }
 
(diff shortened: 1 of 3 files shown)


lib/public/views/Runs/Runs.js
```diff
@@ -196,12 +196,37 @@ export default class Overview extends Observable {
...
      * @returns {Object} The Filtered runs data
      */
-    async getFilteredRuns(runs) {
-        const selectedRuns = runs.filter((run) => this.getSelectedRunNumbers().includes(run.runNumber));
-        const filteredRuns = selectedRuns.map((selectedRun) => pick(selectedRun, this.getSelectedRunsFields()));
+    async getFilteredRuns() {
+        this.runs = RemoteData.loading();
+        this.notify();
+
+        const runIds = this.getSelectedRunNumbers().split(',');
```
graduta, line 205: For `runIds` we know that we expect only valid numbers.
I would recommend to have an extra check in here that the values parsed are indeed numbers and no empty values are passed to the server.
If there is an empty value, it should be dropped from the list of Run Numbers.
If there are no valid numbers in the input, an error message should be printed.

e.g. If by mistake the user inputs the value: `2, 3, ,4,7`, this will generate an empty line in the CSV format

lib/public/views/Runs/Runs.js
```diff
@@ -196,12 +196,37 @@ export default class Overview extends Observable {
...
+        const foundRuns = [];
+        let response;
```
graduta, line 207: `response` and `result` are used only within the `for` loop. 
I would recommend to use them only within the `for` loop scope.

lib/public/views/Runs/Runs.js
```diff
@@ -196,12 +196,37 @@ export default class Overview extends Observable {
...
+        let result;
+
+        // Loops through every run id that has been filled in and retrieves them from the database
+        for (let i = 0; i < runIds.length; i++) {
```
graduta, line 211: Usage of `1 letter` variables is discouraged.
I would recommend you adapt the `for loop` to make use of either:
* [Array.map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) which will retrieve you an array already, perfect for your use case of `foundRuns` and you can also chain it with `filteredRuns` from the end of the method
* [Array.forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach)

lib/public/views/Runs/Runs.js
```diff
@@ -212,9 +237,11 @@ export default class Overview extends Observable {
      * @return {void}
      */
     async createRunsExport(content, fileName) {
-        this.getSelectedExportType() == 'CSV' ?
-            createCSVExport(content, `${fileName}.csv`, 'text/csv;charset=utf-8;')
-            : createJSONExport(content, `${fileName}.json`, 'application/json');
+        if (content.length !== 0) {
```
graduta, line 240: It looks like the user can still input some white spaces and pass this check. 
It would be better to check if the content contains a valid list of numbers

graduta (changes requested):
Good job on identifying the bug due to which only existing runs were used!
A few more changes and this pull request will be good to merge.
Also, please update the branch with the latest changes from master as well

lib/public/views/Runs/Runs.js
Sandervdk: Array.map and Array.forEach are both synchronous functions, I tried running it like this and a couple other similar variations:
- await runIds.map(async runID => { ... })

But they all returned lists of promises.

So I'll change the for loop into a for-of loop which can be used async and give the variable a proper name.

graduta: Indeed, it will return a list of promises. 

Just as a learning exercise you could probably make use of `Promise.all()` in combination with the map. There is an nice example here: https://advancedweb.hu/how-to-use-async-functions-with-array-map-in-javascript/

No need to update the code thought, the `for of` is good enough.

lib/public/views/Runs/Runs.js
```diff
@@ -196,13 +196,47 @@ export default class Overview extends Observable {
 
     /**
      * Filter runs based on the selected run numbers and fields to prepare it for export
```
graduta, line 198: One last small change.
Would you mind updating the method documentation with the new functionality?

graduta (approved):

graduta (dismissed):

graduta (commented):
Ok, so there is a small issue here.

The user is asked to input `runNumber` while the query to the server expects `runIds`
These are different and currently Bookkeeping does not accept `run  numbers` as parameter. There is a ticket in progress for this. 

There is nothing wrong with this PR it just needs to keep this ticket in progress until the Bookkeeping API accepts RUN as parameter

PR #457, opened 2021-12-03
[O2B-444] - wrap run table cells

4libutt:
wrapped table cells, added .toString to detectors and placed detectors second in the run order


4libutt:
I also changed the order of the run detail page, so that detectors are second in the table there as well for a uniform look and feel. is that okay?

graduta (changes requested):
Hi @4libutt ,
When I use `Inspect` from browser and modify the author to a really long title, the cell does not look wrapped to me.  
Would you mind having a further look?
Screenshot attached: 
image

4libutt:
I figured out that because of the `cell-f` or `cell-fm` which have a width in percentages instead of the max-width in rem like the other cell classes, that the `w-wrapped` isn't applying as intended. if I change the cell classes to `cell-l` or `cell-m` the table on log entries will look like the image below. Is that okay?

Screenshot 2021-12-10 at 09 52 15
Screenshot 2021-12-10 at 09 57 04

graduta:
> I figured out that because of the `cell-f` or `cell-fm` which have a width in percentages instead of the max-width in rem like the other cell classes, that the `w-wrapped` isn't applying as intended. if I change the cell classes to `cell-l` or `cell-m` the table on log entries will look like the image below. Is that okay?
> 
> Screenshot 2021-12-10 at 09 52 15
> 
> Screenshot 2021-12-10 at 09 57 04

Hi @4libutt ,
I had a chat with the users and all of them had a large set of different widths for their monitor. For example, the run coordinator is using ultra-wide monitors while the remote shifters use 15" laptops. 

Because of this, I would suggest we use a `%` size for all the rows. Instead of using `cell-f` and `cell-fm`, I would suggest you find a combination with the `w-5, w-10, etc.` for all columns based on their stored values which would sum up to 100%
In this way the size of the cells will adapt based on the users devices.

graduta (approved):

PR #463, opened 2021-12-09
[O2B-466] Pages should display table on whole page

Sandervdk:
Added some UI changes to make the tables match for every page
code now loads more logs, runs and tags to make the tables take up the majority of the page

#### I have a JIRA ticket
- [X] PR labels are selected


lib/public/Model.js
```diff
@@ -86,10 +86,18 @@ export default class Model extends Observable {
         // Reset infinite scroll to prevent scroll listening on pages where it is disabled or not needed
         window.onscroll = null;
 
+        // Calculates the number of rows which should be visible on the page
+        const TABLEROW_HEIGHT = 45;
+        // Estimate as to how much height the navbar and other elements take up (based on the tagsOverview page)
+        const PAGE_USED_HEIGHT = 350;
```
graduta, line 92: Currently it looks like the tables will not resize if the page size is changed.
These calculations should be part of the virtual node of the table so that they get recalculated on each page resize

graduta (changes requested):

graduta (changes requested):
Hi @Sandervdk ,

Changes look great for the `Home` page! Good job!

There is an issues for `Runs` and `Logs` page that once is fixed, I think we can merge this pull request. 
Buttons which allow the user to select the number of `logs`/`runs` per page does not work anymore.

Sandervdk:
Hey @graduta ,

The buttons should work properly again now, this issue also existed in the tags overview page where it should also be fixed now.

graduta (changes requested):
Hi @Sandervdk !
Thank you for the update. I will check the functionality of the buttons. 

In the meantime, I believe tests should be updated as well. 
A scenario from the top of my head is the following:
* puppeteer opens the web page in headless mode but still has a certain size(height and width)
* by looking at the failing test from now it looks like there are only 5 logs displayed per page now rather than the expected 10; https://github.com/AliceO2Group/Bookkeeping/runs/4869004592?check_suite_focus=true#step:6:1633 
* I think the easiest is to increase the size of the puppeteer page to match the default size of a user screen

graduta (approved):

PR #465, opened 2021-12-20
O2B-456: Filter hide and or radio buttons when no tags are present

whdegraaf:
#### I have a JIRA ticket
- [X] PR labels are selected

Modified RunsFilter/tags.js to have the same change as LogsFilter/tags.js upon request.


graduta (approved):
Thank you for adding the check on the Tags page as well!

PR #482, opened 2022-01-24
[O2B-463] Parent title added to titleless replies

Sandervdk:
Added the title of the parent log to the reply, prefased with a "Re: ", to replies with no actual title

#### I have a JIRA ticket
- [X] PR labels are selected


graduta (changes requested):
Hi @Sandervdk ,

Thank you for your contribution!

The changes look good, now the only thing to remember is to add tests. By adding tests, we can ensure the reliability of Bookkeeping!

Following your changes I expect tests that need to be updated/added are in the following places:
* Add/Update the tests which are creating new logs. e.g `CreateLogUseCase.test.js` - https://github.com/AliceO2Group/Bookkeeping/blob/master/test/usecases/log/CreateLogUseCase.test.js#L79 
* Add a new test in `create.test.js` which should create a log as a reply

graduta:
Moreover, you should update the `ActiveColumns` for Logs so that `title` is not mapped to a new string if it is missing

PR #486, opened 2022-01-25
[O2B-281] Sequelize migration for tags table to add notification data

awegrzyn:


awegrzyn:
@vascobarroso

vascobarroso:
@awegrzyn this looks quite nice! Just for my understanding: does this mean we can use `sequelize` for DB schema evolution ?

awegrzyn:
@vascobarroso I thought this is how it's been done? Am I wrong? If so, do you know how database changes has been handled so far?

vascobarroso:
@awegrzyn as far as I know, all DB changes so far have been handled manually, so this is a good step forward.

awegrzyn:
@graduta I've added changes to model as well, could you just lemme know if I shall provides other properties then `type` as well? (eg. `defaultValue` or `allowNull`) This is not the case in other model files but would make sense.

graduta:
Hi @awegrzyn ,

There are a few things to tackle to validate the PR:
- lint process is failing, please have a look on that
- I believe you will need changes to `TagAdapter.js` so that the `toEntity()` method returns the new fields as well (if you do not plan to query this information through Bookkeeping than it is fine)
- I think `defaultValue` or `allowNull` should be added in both model and db migration file; This is because I think there are still plans to give access to the db to other tools;
- If they are added to db schema then we need to update existing tags as well during deployment

graduta (dismissed):

graduta (approved):

PR #500, opened 2022-02-11
[O2B-422] Allow to set counters to zero on FLP update

awegrzyn:


lib/domain/dtos/UpdateFlpDto.js
```diff
@@ -17,6 +17,7 @@ const EntityIdDto = require('./EntityIdDto');
 const OptionalPositiveIntegerDto = Joi.number()
     .integer()
     .positive()
```
graduta, line 19: You can replace `.positive().allow(0)` with `min(0)`

awegrzyn: Isn't it the same?

graduta: Yes, but in one line 😄 
Not a blocker comment though

graduta (approved):

PR #514, opened 2022-02-16
[O2B-484] Fix C++ integration by bringing back runQuality to EndRun DTO

awegrzyn:


lib/domain/dtos/EndRunDto.js
```diff
@@ -16,6 +16,7 @@ const Joi = require('joi');
 const BodyDto = Joi.object({
     timeO2End: Joi.date(),
     timeTrgEnd: Joi.date(),
+    runQuality: Joi.string().valid('good', 'bad', 'test', '').optional(),
```
graduta, line 19: `runQuality` should not be brought back and also `empty string` should not be accepted.
We can keep it here for validation and backwards compatibility but it should not be added in the `EndRunUseCase`

graduta (changes requested):

graduta (approved):

PR #523, opened 2022-03-14
[O2B-500] Display of email groups and mattermost channels in tag overview

JikHuijberts:
#### I have a JIRA ticket
- [x] PR labels are selected
Other boxes are not checked due to JIRA permissions on my side.

When hovering over the items with the mouse will now be a clickable pointer. Furthermore the Mattermost and email fields are now shown in the tag overview.


graduta (changes requested):
If changes are applied to the `openapi` files then the libraries need to be regenerated as well. 
I think it is better if we keep the `openapi` changes for a separate pull request at the end of this epic. 
In this way we update the API and regenerate the libraries as per the documentation. e.g. documentation for go client-library https://github.com/AliceO2Group/Bookkeeping/tree/master/go-api-client#usage

graduta (dismissed):

PR #526, opened 2022-03-15
[O2B-421] Trigger notification on log creation

awegrzyn:


lib/config/notification.js
```diff
@@ -0,0 +1,19 @@
...
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+// Default configuration
+const brokers = [process.env?.NOTIFICATION_BROKER] || null;
+
+module.exports = {
+    brokers,
```
graduta, line 18: I believe the notification service from web-ui expects an array;
Could you confirm that the process.env will return an `Array` and not a `String`?

lib/server/controllers/logs.controller.js
```diff
@@ -54,6 +58,11 @@ const createLog = async (request, response, next) => {
         response.status(Number(error.status)).json({ errors: [error] });
     } else {
         response.status(201).json({ data: result });
+        const url = `${request.protocol}://${request.get('host')}?page=log-detail&id=${result.id}`;
+        const tags = result.tags.map((tag) => tag.text);
+        if (notification && notification.isConfigured()) {
```
graduta, line 63: Here you could use the same syntax you used above
```suggestion
        if (notification?.isConfigured()) {
```

graduta (changes requested):

lib/config/notification.js
awegrzyn: Nope, it will not, but that's why `[` and `]` were added.

graduta: Ah, ok. I missed the brackets as I was thinking we are expecting multiple brokers separated by commas

graduta (dismissed):

graduta (approved):

PR #530, opened 2022-03-22
[O2B-508] When creating logs double run numbers are filtered in the backend

JikHuijberts:
#### I have a JIRA ticket
- [x] PR labels are selected

There was a bug where when duplicate run numbers are inserted, an internal server error would appear and not save the data. This is now fixed and tested with e2e and use case tests.

### DIFF
--- a/test/usecases/log/CreateLogUseCase.test.js
+++ b/test/usecases/log/CreateLogUseCase.test.js
@@ -94,4 +94,24 @@ module.exports = () => {
         expect(result.title).to.equal(expectedTitle);
         expect(result.parentLogId).to.equal(expectedParentLogId);
     });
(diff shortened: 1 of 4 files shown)


test/usecases/log/CreateLogUseCase.test.js
```diff
@@ -94,4 +94,24 @@ module.exports = () => {
         expect(result.title).to.equal(expectedTitle);
         expect(result.parentLogId).to.equal(expectedParentLogId);
     });
+    it('should create a new Log and filter the title ', async () =>{
```
graduta, line 97: I believe the test description is not in line with what is actually tested. Could you please update it?

graduta (changes requested):

graduta (approved):

PR #547, opened 2022-04-01
[O2B-504] Create tag and add email and mattermost

JikHuijberts:
#### I have a JIRA ticket
- [x] PR labels are selected


lib/public/views/Tags/Tags.js
```diff
@@ -75,6 +75,24 @@ class TagModel extends Observable {
         return this.text;
     }
 
+    /**
+     * Getter for email
+     *
+     * @returns {String} The text currently inserted in the tag creation screen.
+     */
+    getEmail() {
+        return this.email;
```
graduta, line 84: Similar to the RUNS and LOGS refactoring, I would suggest we start using the standard convention for JS getters and setters

lib/server/controllers/tags.controller.js
```diff
@@ -41,7 +41,17 @@ const createTag = async (request, response, next) => {
     if (!value) {
         return;
     }
-
+    if ((value?.body?.email || value?.body?.mattermost) &&
+            !request?.session?.access?.includes('admin')) {
+        response.status(403).json({ errors: [
+            {
+                status: '403',
+                title: 'Forbidden',
```
graduta, line 49: This created a bit of a problem recently as the error message was not very informative. I would suggest we give a bite more information. Something in line with
`Non-Admin users are not allowed to provide email and mattermost details for tags"

graduta (changes requested):

lib/public/views/Tags/Create/index.js
```diff
@@ -42,7 +43,21 @@ const createTagScreen = (model) => {
                 value: text,
                 oninput: (e) => model.tags.setText(e.target.value),
             }, text),
+            model.session?.access?.includes('admin') && [
+                h('label.form-check-label.f4.mt2', { for: 'mattermost' }, 'Mattermost channels'),
```
graduta, line 47: Could we add a descriptive label above "Mattermost" and "Email" Fields so that is clear to the users what are their tags going to do. Something similar to:
```
Tags should inform below channels and groups on log creation:
Mattermost channels:
<mattermost-field>
Email Groups:
<email-groups-field>
```

lib/public/views/Tags/Tags.js
```diff
@@ -292,7 +333,9 @@ class TagModel extends Observable {
      * @returns {undefined}
      */
     clearEditor() {
-        this.text = '';
+        this._text = '';
+        this._mattermost = '',
+        this._email = '',
```
graduta, line 338: As we change the state we should also notify about it, especially as one of the objects is a RemoteData one

graduta (approved):

PR #567, opened 2022-05-02
[O2B-544] Next and previous buttons in the runs page

JikHuijberts:
The next and previous buttons show whenever there is a next value or previous value. It also works on runs with a filter applied.
 
#### I have a JIRA ticket
- [x] PR labels are selected

### DIFF
--- a/lib/public/views/Runs/Details/index.js
+++ b/lib/public/views/Runs/Details/index.js
@@ -28,6 +28,10 @@ import editTagsPanel from './editTagsPanel.js';
  * @return {Object} A key-value collection of all relevant fields
  */
 const runDetails = (model) => {
+    let prevId = null;
+    let nextId = null;
+
+    const runs = model.runs.getRuns();
     if (!model.router.params.id) {
         model.router.go('?page=run-overview');
         return;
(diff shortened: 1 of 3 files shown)


lib/public/views/Runs/Runs.js
```diff
@@ -40,9 +40,12 @@ export default class Overview extends Observable {
 
     /**
      * Retrieve every relevant run from the API
+     * @param {boolean} add Selects if you want to add the runs to the existing list
      * @returns {undefined} Injects the data object with the response data
      */
-    async fetchAllRuns() {
+    async fetchAllRuns(add = false) {
+        const runs_c = this.runs;
+        this.runs = RemoteData.loading();
```
graduta, line 48: We need to use `notify` if you want for `this.runs` to display the change
```
this.runs = RemoteData.loading();
this.notify();
```

lib/public/views/Runs/Details/index.js
```diff
@@ -115,6 +131,14 @@ const runDetails = (model) => {
                     h('.w-40.ph1', runDetail(model, data.payload)),
                     model.runs.isEditModeEnabled && editTagsPanel(model),
                 ]),
+                h('.flex-row.w-100.justify-center', [
+                    prevId && h('button.btn.btn-primary.m2#prev-run', {
+                        onclick: () => model.router.go(`/?page=run-detail&id=${prevId}`),
+                    }, 'Prev'),
```
graduta, line 137: I think the logic here should be the other way around. 
`next` for latest runs
`prev` for previous runs
Locally while testing I have seen that `next` took me from Run7 to Run6

graduta (changes requested):

lib/public/views/Runs/Details/index.js
JikHuijberts: Prev and next are switched now

graduta (changes requested):
I updated the `run-details.test` and fixed the issue. There are still issues with the `run-overview` tests. As far as I can tell the date is introduced in the other way around, at least that it what appears in the logs.

Could you please have a look at the rest of the comments and we fix the tests afterwards

lib/public/views/Runs/Details/index.js
```diff
@@ -37,7 +41,19 @@ const runDetails = (model) => {
         model.router.go(targetURL(model, 'panel', 'logs'), true);
         return;
     }
-
+    if (!runs.isLoading() && runs.isNotAsked()) {
```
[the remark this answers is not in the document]
JikHuijberts: There is a not operator in front of it. This prevents fetches while the data is still loading but a `notify` is triggered.

PR #572, opened 2022-05-05
Docs updated, envs smaller table and shows title on hover

JikHuijberts:
#### I DON'T have JIRA ticket
- [X] explain what this PR does
- [X] if it is a new feature, explain how you plan to use it
- [X] tests are added
- [X] documentation was updated or added

ActiveColumns on tables now has a field `title: true/false` that adds the title property. The title property has the text as value, this is also compatible with the format field.

Furthermore the database.md is updated to include the **Environments** table with the fields.


graduta (approved):
Thank you for the quick action!

PR #615, opened 2022-06-07
[O2B-511] Run filters should accept operators

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

### DIFF
--- a/lib/public/components/Filters/components/amountFilter.js
+++ b/lib/public/components/Filters/components/amountFilter.js
@@ -1,6 +1,3 @@
-import { comparisonOperatorFilter } from './comparisonOperatorFilter.js';
-import { h } from '/js/src/index.js';
-
 /**
  * @license
  * Copyright CERN and copyright holders of ALICE O2. This software is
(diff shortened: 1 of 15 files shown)


lib/public/components/Filters/components/amountFilter.js
```diff
@@ -0,0 +1,54 @@
+import { comparisonOperatorFilter } from './comparisonOperatorFilter.js';
+import { h } from '/js/src/index.js';
+
+/**
```
graduta, line 4: I believe the import should come after the `licence` tag and comment, right?

graduta (changes requested):
I found an interesting user scenario which it is not desirable.
1. Open Run-Overview Page
2. Open Filters
3. Select an operator before adding a value for a numerical filter;
This will transition the page to an error state due to `Invalid Attribute: "query.filter.nDetectors.limit" must be a number`

martinboulais:
> I found an interesting user scenario which it is not desirable.
> 
>     1. Open Run-Overview Page
> 
>     2. Open Filters
> 
>     3. Select an operator before adding a value for a numerical filter;
>        This will transition the page to an error state due to `Invalid Attribute: "query.filter.nDetectors.limit" must be a number`

I fixed it

graduta (approved):

PR #627, opened 2022-06-22
[O2B-601] Add run duration column and filter

martinboulais:
#### I have a JIRA ticket
- [x] PR labels are selected
- [ ] `CHANGELOG.md` and `database/CHANGELOG.md` files were updated

### DIFF
--- a/lib/public/components/Filters/RunsFilter/durationFilter.js
+++ b/lib/public/components/Filters/RunsFilter/durationFilter.js
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright CERN and copyright holders of ALICE O2. This software is
+ * distributed under the terms of the GNU General Public License v3 (GPL
+ * Version 3), copied verbatim in the file "COPYING".
+ *
+ * See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { amountFilter } from '../components/amountFilter.js';
+
+/**
+ * Returns the run duration filter component
+ *
+ * @param {Object} model The global model object
+ *
+ * @return {vnode} the duration filter
+ */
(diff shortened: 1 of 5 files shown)


lib/public/components/Filters/RunsFilter/durationFilter.js
```diff
@@ -0,0 +1,36 @@
...
+ * Returns the run duration filter component
+ *
+ * @param {Object} model The global model object
+ *
+ * @return {vnode} the duration filter
+ */
+export const durationFilter = (model) => amountFilter(
+    model.runs.runDurationFilter,
+    (filter) => {
+        model.runs.runDurationFilter = filter;
```
graduta, line 26: I think it would be worth adding a simple check here not only for `null` but for the type of input as well. 

While it is true, the input box will clear the value if it is not a digit, based on the network panel we still query again for the `runs` but with no filter value. 

Thus, to avoid extra calls we could check for `typeof` if the setter does not fetch for new runs if value is NaN?

lib/public/views/Runs/ActiveColumns/index.js
```diff
@@ -92,6 +93,21 @@ const activeColumns = (model) => ({
         size: 'w-10 f6 w-wrapped',
         format: (timestamp) => formatTimestamp(timestamp, false),
     },
+    runDuration: {
+        name: 'Duration',
```
graduta, line 97: Even though it may be obvious, users would like to name it `Run Duration`

lib/public/components/Filters/RunsFilter/durationFilter.js
martinboulais: The problem here is that I need to send a request without filter when the limit is empty. The problem is that some navigators send null as input value when entering not numerical value in input type number. The solution I have is to send request only if the value is not the same as the previous one (see fix)

lib/public/views/Runs/ActiveColumns/index.js
martinboulais: Done

graduta (approved):

graduta (approved):

PR #640, opened 2022-07-04
[O2B-631] Create Run & Update Run OpenAPI

JikHuijberts:
#### I have a JIRA ticket
- [] issue has "Fix version" assigned
- [X] PR labels are selected
- [ ] `CHANGELOG.md` and `database/CHANGELOG.md` files were updated

For go
Extra fields are **added** for updating the run:
- timeO2Start
- timeTrgStart
- trgGlobalRunEnabled 
- trgEnabled
- pdpConfigOption
- pdpTopologyDescriptionLibraryFile
- tfbDdMode

Two fields are **removed** for the create run function:
- timeO2Start
- timeTrg


lib/database/models/run.js
```diff
@@ -104,6 +104,24 @@ module.exports = (sequelize) => {
         aliceDipolePolarity: {
             type: Sequelize.CHAR(32),
         },
+        trgGlobalRunEnabled: {
+            type: Sequelize.BOOLEAN,
+        },
+        trgEnabled: {
+            type: Sequelize.BOOLEAN,
+        },
```
martinboulais, line 112: You put default null in migration but not here, is it normal?

lib/domain/dtos/EndRunDto.js
```diff
@@ -14,8 +14,15 @@
 const Joi = require('joi');
 
 const BodyDto = Joi.object({
-    timeO2End: Joi.date(),
-    timeTrgEnd: Joi.date(),
+    timeO2Start: Joi.date().allow(-1).optional(),
+    timeTrgStart: Joi.date().allow(-1).optional(),
+    timeO2End: Joi.date().allow(-1).optional(),
+    timeTrgEnd: Joi.date().allow(-1).optional(),
```
martinboulais, line 20: Can we think about another solution than -1? Sending nullable timestamp for example. 
Because here the -1 seems to come out of nowhere when you do not know the context, and actually negative timestamps may have a meaning (probably not in this application for sure)
I really think data must be self explicit, and using a not null value to represent null is really error prone

lib/usecases/run/EndRunUseCase.js
```diff
@@ -50,8 +61,32 @@ class EndRunUseCase {
...
+                }
+                if (timeO2Start && timeO2Start > 0) {
+                    runEntity.timeO2Start = timeO2Start;
+                }
+                if (timeTrgEnd && timeTrgEnd > 0) {
+                    runEntity.timeTrgEnd = timeTrgEnd;
+                }
+                if (timeTrgStart && timeTrgStart > 0) {
+                    runEntity.timeTrgStart = timeTrgStart;
+                }
```
martinboulais, line 78: I do not see the purpose of the `> 0` check. Is it because of the `-1`? Then it is even less obvious, maybe we should at least put a constant and test it exactly

lib/usecases/run/EndRunUseCase.js
```diff
@@ -50,8 +61,32 @@ class EndRunUseCase {
...
+                runEntity.trgEnabled = trgEnabled;
+                if (pdpConfigOption) {
+                    runEntity.pdpConfigOption = pdpConfigOption;
+                }
+                if (pdpTopologyDescriptionLibraryFile) {
+                    runEntity.pdpTopologyDescriptionLibraryFile = pdpTopologyDescriptionLibraryFile;
+                }
+                if (tfbDdMode) {
+                    runEntity.tfbDdMode = tfbDdMode;
+                }
```
martinboulais, line 89: Maybe we can do it more straightforward it, using `spread syntax` or extract a function, because we will probably do the same many times (patch existing object) :
```js
runEntity = {
    ...runEntity,
    ...body.runQuality || runEntity.runQuality
    // etc
}
for the fields that do not need supplementary checks ?
```

lib/database/migrations/20220704125109-run-techincal-board-support-fields.js
```diff
@@ -0,0 +1,39 @@
...
+            allowNull: true,
+            default: null,
+        }, { transaction: t }),
+        queryInterface.addColumn('runs', 'pdp_topology_description_library_file', {
+            type: Sequelize.DataTypes.STRING,
+            allowNull: true,
+            default: null,
+        }, { transaction: t }),
+        queryInterface.addColumn('runs', 'tfb_dd_mode', {
+            type: Sequelize.DataTypes.CHAR(64),
```
martinboulais, line 26: For my information: why not using `STRING(64)`?

lib/domain/dtos/EndRunDto.js
JikHuijberts: The go API can only send int64 types so we are restricted to numbers. Furthermore, we do not want to save dates before 2020 so that's why the decision was made to use -1 as a null. For the users that use the API the values still need to accept non-existing/empty values so that's why there is still an optional option with the validation. Small example [here](https://go.dev/play/p/w8Qdhi1fTMX)

lib/database/models/run.js
JikHuijberts: Yes, this is for the already existing values. Within the dto, it is already handled to be false when there are new values put in (because again of the go API limits).

lib/domain/dtos/EndRunDto.js
JikHuijberts: Discussed this with george as well we are going to change it to be handled within the `BookkeepingApi.go`

lib/usecases/run/EndRunUseCase.js
JikHuijberts: Opted for using `runEntity.update(body)`

PR #645, opened 2022-07-07
[O2B-639] Add new info from Core to Run-Details

JikHuijberts:
#### I have a JIRA ticket
- [x] PR labels are selected
- [x] `CHANGELOG.md` and `database/CHANGELOG.md` files were updated

Run details now show the extra fields stated below:
- Trigger Global Run Enabled;
- Trigger Enabled;
- PDP Configuration Option;
- PDP Topology Description Library File;
- TFB DD Mode;

And created a default value that can be set for `null` values.


lib/public/utilities/formatTimestamp.js
```diff
@@ -43,3 +43,21 @@ export const formatTimestamp = (timestamp, inline = true) => {
...
+ */
+export const formatBoolean = (boolean) => {
+    if (boolean != null) {
+        if (boolean) {
+            return 'On';
+        }
+        return 'Off';
+    }
+    return '-';
+};
```
martinboulais, line 63: Can you please put this in its own file?
I think in general the best is one exported function => one file, with the same name as the exported function

lib/public/components/RunDetail/index.js
```diff
@@ -122,19 +142,19 @@ const activeFields = (model) => ({
         name: 'Data Distribution (FLP)',
         visible: true,
         size: 'cell-s',
-        format: (boolean) => boolean ? 'On' : 'Off',
+        format: (boolean) => formatBoolean(boolean),
```
martinboulais, line 145: simply give the function reference as property, no need to wrap in a function (same for the others)

martinboulais (changes requested):

lib/public/components/Detail/index.js
```diff
@@ -19,14 +19,25 @@ import { h } from '/js/src/index.js';
...
+    key, {
+        name,
+        format,
+    },
+]) => h(
+    `.flex-row.justify-between#${selectorName}-${key}`,
+    h('b', `${name}:`),
+    (()=> {
+        if (format) {
+            return format(post[key]);
```
martinboulais, line 35:
```suggestion
            return format(post[key], post);
```

martinboulais (changes requested):

lib/public/components/Detail/index.js
JikHuijberts: Fixed it thanks for the tip!

lib/public/components/RunDetail/index.js
JikHuijberts: done!

lib/public/utilities/formatTimestamp.js
JikHuijberts: Done!

martinboulais (approved):

PR #741, opened 2022-09-26
[O2B-709] Display timestamps as date or raw timestamps

martinboulais:
#### I have a JIRA ticket
- [x] PR labels are selected
- [x] `CHANGELOG.md` and `database/CHANGELOG.md` files were updated


lib/public/utilities/userPreferencesStore.js
```diff
@@ -0,0 +1,35 @@
...
+        return this._rawTimestamps;
+    }
+
+    /**
+     * Set if the timestamp must be defined as raw timestamps or formatted as date
+     *
+     * @param {boolean} value the new preference
+     */
+    set rawTimestamps(value) {
+        this._rawTimestamps = value;
```
graduta, line 31: In order to catch non-boolean values, it might be worth to use `Boolean(value)` here as well? Similar to the one you used for https://github.com/AliceO2Group/Bookkeeping/blob/main/lib/public/views/Tags/Tags.js#L302

lib/public/components/NavBar/index.js
```diff
@@ -90,14 +93,29 @@ const navBar = (model) => {
...
+                            id: 'preferences-raw-timestamps',
+                            onclick: () => {
+                                userPreferencesStore.rawTimestamps = !userPreferencesStore.rawTimestamps;
+                            },
+                            type: 'checkbox',
+                            checked: userPreferencesStore.rawTimestamps,
+                        }),
+                        h('label', {
+                            for: 'preferences-raw-timestamps',
+                        }, 'Display raw timesamps'),
```
graduta, line 108: `timestamps`

graduta (changes requested):
There is a small issue in which:
- enabling `timestamps` only
- navigate to a run (run-details) from run-overview
Time and Date are now reverted to using normal formatting and preference is not preserved. I think this is due to the wrong way the navigation is handled from the overview page. 

While this is a separate bug, I think it would be good to take care of it here.

lib/public/components/NavBar/index.js
martinboulais: Fixed, both

graduta (approved):

PR #756, opened 2022-09-30
[O2B-735] Improve pagination

martinboulais:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users
- Overview pagination has been uniformized on every pages


lib/public/components/Pagination/amountSelector.js
```diff
@@ -34,61 +36,84 @@ function isPerPageAmountValid(amount) {
...
     oninput: ({ target }) => {
-        model.setCustomPerPage(target.value);
-        model.notify();
+        const customItemsPerPage = parseInt(target.value, 10);
+        if (!isNaN(customItemsPerPage)) {
+            paginationModel.customItemsPerPage = customItemsPerPage;
+        }
     },
-    onblur: ({ target }) => isPerPageAmountValid(target.value) && onblur(target.value),
+    onblur: ({ target }) => isPerPageAmountValid(target.value) && onCustomChoiceConfirm(target.value),
```
graduta, line 55: I think it would be great to add an event listener for pressing `Enter` as well which would then trigger the choice confirmation and hide the selector. 
What do you think?

graduta (changes requested):
It looks like there are still a few bugs (maybe not entirely related to refactoring of the pagination):
- LHC Fill overview, when selecting Infinite and going to the bottom of the page to request more date will automatically go back to top of the table
- all pages with pagination, if you start from page 4 and select Infinite, it will not not display pages from 1 to 4 anymore

lib/public/components/Pagination/amountSelector.js
martinboulais: >     * LHC Fill overview, when selecting Infinite and going to the bottom of the page to request more date will automatically go back to top of the table

Probably a bug I added while keeping the branch up to date, I'll have a look

>     * all pages with pagination, if you start from page 4 and select Infinite, it will not not display pages from 1 to 4 anymore

If I am not mistaken this is a bug that existed before, I tried to keep this PR as simple as possible (it already has quite a lot of changes)

martinboulais: > I think it would be great to add an event listener for pressing `Enter` as well which would then trigger the choice confirmation and hide the selector. What do you think?

I agree with you. Should I add it in this PR?

martinboulais: The first one is fixed, I'll create the tickets

graduta (approved):

graduta (approved):

PR #769, opened 2022-10-11
[O2B-741] Runs must overlap with stable beam to be considered as physics 

martinboulais:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Restrict physics runs to runs that overlap with a stable beam


graduta (changes requested):
While this fixes the issue for RunDetails and RunOverview, I think we still need to address the LHC Fill Details page as well which still does not look at stable beam time versus the run time. e.g fill 8223

graduta (approved):

PR #771, opened 2022-10-11
[O2B-749] Improve run quality update logging

martinboulais:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Log only run quality change from good to bad and the other way around
- Add information to the log created when run quality change such as user and date

### DIFF
--- a/lib/server/services/user/getUser.js
+++ b/lib/server/services/user/getUser.js
@@ -13,19 +13,26 @@
 
 const { utilities: { QueryBuilder } } = require('../../../database');
 const UserRepository = require('../../../database/repositories/UserRepository.js');
+const { BadParameterError } = require('../../errors/BadParameterError.js');
 
 /**
- * Find and return an user model by its id
+ * Find and return a user model by its id
  *
- * @param {string} userId the id of the user to find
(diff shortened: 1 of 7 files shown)


lib/server/services/user/getUser.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param {string} userId the id of the user to find
+ * @param {function|null} qbConfiguration function called with the user find query builder as parameter to add specific configuration to
+ *     the query
+ * @return {Promise<SequelizeUser|null>} the user found or null
+ */
+exports.getUser = (userId, qbConfiguration = null) => {
+    const queryBuilder = new QueryBuilder();
+
+    queryBuilder.where('id').is(userId);
```
graduta, line 28: I believe the `id` fields in the `users` table is the auto-incremented one.
In this scenario, we expect the `external_id` to be received from session DTO and we should query based on that

graduta (changes requested):

lib/server/services/user/getUser.js
martinboulais: Changed to allow for externalId too

graduta (approved):

PR #787, opened 2022-10-28
[O2B-607] Add possibility to archive tags

martinboulais:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Tags can now be marked as archived by administrators, and those tags can not be linked to new runs or logs


test/e2e/logs.test.js
```diff
@@ -922,7 +900,7 @@ module.exports = () => {
                 });
         });
 
-        it('should return 400 if at least one non-existent tag was provided', (done) => {
+        it('should return 400 if at least one non-existent or archived tag tag was provided', (done) => {
```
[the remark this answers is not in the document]
martinboulais: Fixed

lib/database/adapters/TagAdapter.js
```diff
@@ -29,13 +29,15 @@ class TagAdapter {
...
-    toEntity({ id, text, email, mattermost, last_edited_name, updatedAt }) {
+    toEntity({ id, text, email, mattermost, last_edited_name, archived, archivedAt, updatedAt }) {
         return {
             id,
             text,
             lastEditedName: last_edited_name,
             email,
             mattermost,
+            archived,
+            archivedAt: archivedAt ? new Date(archivedAt).getTime() : null,
```
graduta, line 40: In the interest of using the sane standard, should this line be written with `||`?

lib/public/views/Tags/Details/index.js
```diff
@@ -60,51 +62,49 @@ const tagDetails = (model) => {
...
+                            'h2.mv2.flex-row',
+                            {
+                                onremove: () => {
+                                    model.tags.clearTagChanges();
+                                    model.tags.clearTag();
+                                },
+                            },
+                            `Tag: ${tag.text}`,
+                        ),
+                        tag.archived ? h('.f4.badge.bg-warning.white', 'Archived') : null,
```
graduta, line 76: I have seen that for the `tag-overview` page the color `grey` is used for displaying the badge, while here it is used `orange` from `warning` but both badges display the same information. 
I think it would be better to use the same color `grey` to be consistent

lib/public/views/Runs/Details/editTagsPanel.js
```diff
@@ -17,13 +17,13 @@ import { remoteDataTagPicker } from '../../../components/tag/remoteDataTagPicker
...
  * a list of potential tags to add to a RUN
- *a button to update the tag selection
+ * a button to update the tag selection
  * @param {Model} model Pass the model to access the defined functions.
  * @return {vnode} virtual node with representation of the panel
  */
 const editTagsPanel = (model) => h(
     '#tags-selection.flex-column.w-30',
-    remoteDataTagPicker(model.tags.getTags(), model.runs.detailsModel.editionTagPickerModel, { outlineSelection: true }),
+    remoteDataTagPicker(model.tags.getActiveTags(), model.runs.detailsModel.editionTagPickerModel, { outlineSelection: true }),
```
graduta, line 26: It would be great if the page for editing the tags of a run displays also in `()` that the tag is archived so that the user knows if they deselect it, they will not be able to add it back to the run

graduta (changes requested):
One scenario that will let the users confused is the following:
- UserA opens the `log-creation` page and fills in the form by selecting tag `A`
- In the meantime, UserB archives the tag `A`
- When the user will click `Post Log`, they will receive the following error message (screenshot below) which says `Tags DPG could not be found` but the tag is there visible in their selection form
In this scenario, it would be great if the error message is something similar to what you added for `Run Update` that the tag cannot be used as it is archived

lib/database/adapters/TagAdapter.js
martinboulais: We do not use archivedAt "as is" so we cannot do that

lib/public/views/Tags/Details/index.js
martinboulais: Done

lib/public/views/Runs/Details/editTagsPanel.js
martinboulais: Done

martinboulais:
> One scenario that will let the users confused is the following:
> 
>     * UserA opens the `log-creation` page and fills in the form by selecting tag `A`
> 
>     * In the meantime, UserB archives the tag `A`
> 
>     * When the user will click `Post Log`, they will receive the following error message (screenshot below) which says `Tags DPG could not be found` but the tag is there visible in their selection form
>       In this scenario, it would be great if the error message is something similar to what you added for `Run Update` that the tag cannot be used as it is archived

I propose "Tags ... could not be found (they may have been archived)", because we may have a mix of not existing tags or archived ones

graduta (approved):

PR #868, opened 2023-02-06
[O2B-813] Create environment history but keep existing functionalities as is

martinboulais:
#### I have a JIRA ticket
- [x] PR labels are selected

- Environment states are now stored in an history table

Changes made to the database:
- The environment status and statusMessage has been moved to a separate table to store the history of its changes


graduta (changes requested):
Very nice starting point for the environment states history. There are just a couple of things that need to be adjusted:
- for the migration of existing environments, it would be good to define also the `STANDBY` state by using the `createdAt` timestamp; This would be extra to the current migration of `DESTROYED` which uses the `updatedAt` timestamp;
- I confirmed with AlIECS and we should proceed with using an enum for the ENV states; List has been added to the Jira Ticket
- For the `EnvironmentService.test.js` I only see succesful scenarios; It would be nice to integrate a few errors thrown scenarios there;

graduta (approved):

PR #885, opened 2023-02-21
[O2B-771] Improve run's duration missing data message

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- A popover is displayed in plase of `*` and `**` for run duration with missing triggers


lib/public/views/Runs/formatRunDuration.js
```diff
@@ -27,55 +28,45 @@ export const formatRunDuration = (run, raw = false) => {
...
+
+    if (timeTrgEnd || timeO2End) {
+        let warningPopover = null;
 
-        if (raw) {
-            return ret.text;
+        if (missingTimeTrgStart && missingTimeTrgEnd) {
+            warningPopover = 'Duration based on o2 start AND stop because of missing trigger information';
+        } else if (missingTimeTrgStart) {
+            warningPopover = 'Duration based on o2 start because of missing trigger information';
```
graduta, line 42: I think in the 2 scenarios of missing only one of the trigger times, the message should be adapted accordingly.
`missing trigger information` when only trigger start or stop is missing is not correct

graduta (changes requested):
I believe the `home` page also needs to be adapted accordingly

lib/public/views/Runs/formatRunDuration.js
martinboulais: Updated

graduta (dismissed):

graduta (approved):

PR #893, opened 2023-02-24
[O2B-649] Fix EOR reason reverting

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- EOR reason is now reverted when reverting run edition

- Cleaned run detail models (edition)
- Refactored run edition

Changes made to the database:
-


graduta (changes requested):
Refactor looks good but there are 2 scenarios which I think need a closer look:
1. Incorrect general run quality 
If a run has the general quality as "bad/test/none".
I click "edit" and the dropdown goes straight to "good". 
If I click save, the run quality will not be saved as "good" but it will stay the same

2. Adding an EOR reason temporary does not dissappear
As a user, if I open run-details and click "edit"
I select an EOR reason and click the `plus` icon. 
I then change my mind and click `revert` on the RUN so that I start from scratch
I click "save" once again and the EOR reason that I had added before, it is still there

graduta (approved):

PR #951, opened 2023-04-18
[O2B-854] Make dropdown sort case-insensible

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- Search inside dropdown selection is now case-insensitive


graduta (changes requested):
Would it be possible to add 1-2 tests for this so that the change is not reverted by mistake in the future?

PR #1000, opened 2023-05-24
[O2B-562] Colored environment statuses

AnoeBanoe:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Feature implemented, environment status now displays the correct color based off the status.

- Added new utility `coloredEnvironmentStatusComponent`, returns the status formatted with correct color. The entire row now has correct color based off status implemented in `environmentOverviewComponent` with function `getRowsClasses`.


lib/public/components/common/table/rows.js
```diff
@@ -32,6 +32,8 @@ const cell = (rowId, column, rowData) => {
     // Flag defining if the text must have a popover displayed at hover containing the full text
     const isBalloon = typeof balloon === 'function' ? balloon(cellData) : Boolean(balloon);
 
+    const errorStatus = rowData.status === 'ERROR';
```
martinboulais, line 35: Now that I see it, I think that this logic `(if rowData === 'ERROR') // add class danger` would better fit in the column configuration, not directly in the table

lib/public/components/common/table/rows.js
```diff
@@ -41,7 +43,7 @@ const cell = (rowId, column, rowData) => {
     }
 
     // Keep value.size for retro-compatibility purpose
-    return h(`td.${column.size} ${column.classes}`, {
+    return h(errorStatus ? `td.${column.size} ${column.classes}.danger` : `td.${column.size} ${column.classes}`, {
```
martinboulais, line 46: As said before, I think this should not be there. But FYI, with conditionals I prefer to limit the conditional branches to the most atomic things, here I would replace
```js
errorStatus ? `td.${column.size} ${column.classes}.danger` : `td.${column.size} ${column.classes}`
```
by
```js
`td.${column.size} ${column.classes}${errorStatus ? '.danger' : ''`
```

To avoid dupplication of `td.${column.size} ${column.classes}`

lib/public/utilities/setCorrectStatusColor.js
```diff
@@ -0,0 +1,36 @@
...
+import { h } from '/js/src/index.js';
+
+/**
+ * Change the color of the status of the environment based off which status
+ *
+ * @param {String} status the status to check, format and return
+ *
+ * @return {vnode} the formatted status
+ */
+const setCorrectStatusColor = (status) => {
```
martinboulais, line 23: if your function is `set...` it should update something, not return a component. I would suggest something like `coloredEnvironmentStatusComponent`.
Plus, when you use a switch to map between a value and another, basically you need to use an array or an object/map:

```js
switch (status) {
    case 'RUNNING':
        return h('.success', status);
    case 'ERROR':
        return h('.danger', status);
    case 'CONFIGURED':
        return h('.warning', status);
    default:
        return h('', status);
}
```
is equivalent to 
```
const statusToCssClassMapping = {
    RUNNING: '.success',
    ERROR: '.danger',
    CONFIGURED: '.warning'
}
// Es11 & node 14
return h(statusToCssClassMapping[status] ?? '', status);
// Older versions
return h(statusToCssClassMapping[status] || '', status);

File: lib/public/utilities/setCorrectStatusColor.js [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+        case 'ERROR':
+            return h('.danger', status);
+        case 'CONFIGURED':
+            return h('.warning', status);
+        default:
+            return h('', status);
+    }
+};
+
+export { setCorrectStatusColor };
```
martinboulais, line 36: Prefer to use `export` in front of `const` so that we directly see what is exported

lib/public/utilities/setCorrectStatusColor.js
```diff
@@ -0,0 +1,36 @@
+/**
```
martinboulais, line 1: Since this function is used only for environments, it should not be in utilities

test/public/defaults.js
```diff
@@ -271,6 +271,32 @@ module.exports.checkColumnBalloon = async (page, rowIndex, columnIndex) => {
...
 
+/**
+ * Check that the fourth cell of the given column displays the correct color depending on the status
+ *
+ * @param {{$: function}} page the puppeteer page
+ * @param {number} rowIndex the index of the row to look for status color
+ * @param {number} columnIndex the index of the column to look for status color
+ * @returns {Promise<Chai.Assertion>} void promise
+ */
+module.exports.checkStatusColor = async (page, rowIndex, columnIndex) => {
```
martinboulais, line 282: I think you should specify that this refers to environment status, in comment and in the function name

martinboulais (changes requested):

lib/public/components/common/table/rows.js
AnoeBanoe: Tried to accomplish this in ```environmentOverviewComponent.js``` but I can't get it to work. Further assistance needed.

lib/public/views/Environments/ColoredEnvironmentStatusComponent.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { h } from '/js/src/index.js';
+
+/**
+ * Change the color of the status of the environment based off which status
```
martinboulais, line 17: I would suggest to update this to something like `Returns a component displaying a colored environment status`

lib/public/views/Environments/ColoredEnvironmentStatusComponent.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param {String} status the status to check, format and return
```
martinboulais, line 19: `String` => `string`, the docblock seems to be outdated? Maybe `the environment status to display`

lib/public/views/Environments/Overview/environmentOverviewComponent.js
```diff
@@ -36,8 +36,18 @@ export const environmentOverviewComponent = (pagination, envs) => {
...
+            table(envs, environmentsActiveColumns, { classes: getRowsClasses }),
             paginationComponent(pagination),
         ]),
     ];
 };
+
+/**
+ * The shows the environment table
+ * @param {Environment} environment the environment to be formatted
+ * @returns {string} CSS class
```
martinboulais, line 48: Please update this docblock

martinboulais (changes requested):

lib/public/components/common/table/rows.js
```diff
@@ -51,7 +51,9 @@ const cell = (rowId, column, rowData) => {
  * @typedef RowsConfiguration
  * @property {function(Array):Object} [callback] a callback called on each row's data, and whose result will be added to the row component's
  *     attributes
- * @property {string|function} [classes] the css classes to apply to each row (separated by a dot)
+ * @property {string} [classes] the css classes to apply to each row (separated by a dot)
+ * @property {function} [classes] the css classes to apply to each row,
+ * returned from a callback function based off given parameters (separated by a dot)
```
martinboulais:
```suggestion
 * @property {string} [classes] the css classes to apply to each row. 
                                If it is a string, use it as is(separated by a dot)
                                If it is a function, call the function with the row as argument and use the result as a class
 * @property {function} [classes] the css classes to apply to each row,
 * returned from a callback function based off given parameters (separated by a dot)
```

martinboulais (approved):

PR #1005, opened 2023-06-01
[O2B-898] Clean lost runs and environments every 30 seconds

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

- A new route has been added to purge lost runs and detectors


lib/application.js
```diff
@@ -54,6 +55,16 @@ class BookkeepingApplication {
...
+
+            // Start background housekeeping
+            setInterval(async () => {
+                try {
+                    await handleLostRunsAndEnvironments();
+                    this.logger.debug('Updated lost runs and environments');
+                } catch (error) {
+                    this.logger.error(`Error while handling lost logs: ${error}`);
+                }
+            }, 15 * 60 * 1000);
```
graduta, line 67: I would say 15 minutes will for sure impact the statistics.
Initially, we agree on 15 seconds but to be on the safe side, perhaps we start with 30 seconds?

lib/server/controllers/environments.controller.js
graduta, line 1: As we are not using a ROUTE anymore, is there a point of these imports in the controller?

lib/server/services/environment/transitionToErrorLostEnvironments.js
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @param {string[]} activeEnvironmentIds the ids of active environments (those environments will not be transitioned to error)
+ * @param {number} modificationTimeWindow the time we have (in MINUTES), after the creation of an environment, to purge them. If an environment
+ *     has been created more than this amount of minutes before NOW, it is NOT purged
+ * @return {Promise<void>} resolve once all inactive environments has been transitioned to error
+ * @deprecated
+ */
+exports.transitionToErrorLostEnvironments = async (activeEnvironmentIds, modificationTimeWindow) => sequelize.query(`
+    INSERT INTO environments_history_items (environment_id, status, status_message)
+    SELECT e.id, 'ERROR', 'Shutdown not received, environment not existing anymore'
```
graduta, line 27: I would rephrase it a bit to something more like: `Shutdown not received but environment does not exist anymore`

graduta (changes requested):
As the PR feature changed to not use an API endpoint anymore, please update the PR title as well.

lib/application.js
martinboulais: I misunderstood, I also thought it was a lot... Let's try with 30 seconds yes

lib/server/controllers/environments.controller.js
martinboulais: No indeed, as the linter says

lib/server/services/environment/transitionToErrorLostEnvironments.js
martinboulais: Done

lib/application.js
```diff
@@ -54,6 +55,16 @@ class BookkeepingApplication {
...
                 await this.gRPCServer.listen(gRPCOrigin);
             }
+
+            // Start background housekeeping
+            setInterval(async () => {
+                try {
+                    await handleLostRunsAndEnvironments();
+                    this.logger.debug('Updated lost runs and environments');
+                } catch (error) {
+                    this.logger.error(`Error while handling lost logs: ${error}`);
```
graduta, line 65: I believe it is not the `logs` that we are handling here

graduta (approved):

PR #1020, opened 2023-06-19
[O2B-846] Remove `Re:` from log title

Ryan-Veerman:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- When replying to a log, the reply title is the same as the parent and is not modifiable

### DIFF
--- a/lib/database/models/dpl/dplprocesstype.js
+++ b/lib/database/models/dpl/dplprocesstype.js
@@ -19,6 +19,7 @@ module.exports = (sequelize) => {
         {
             id: {
                 allowNull: false,
+                autoIncrement: true,
                 primaryKey: true,
                 type: Sequelize.CHAR,
             },
--- a/lib/database/repositories/Repository.js
+++ b/lib/database/repositories/Repository.js
@@ -60,12 +60,15 @@ class Repository {
     /**
      * Returns a specific entity.
      *
-     * @param {QueryBuilder} queryBuilder The QueryBuilder to use.
+     * @param {Object|QueryBuilder} findQuery the find query (see sequelize findAll options) or a find query builder
      * @returns {Promise<Object>|null} Promise object representing the full mock data
      */
-    async findOne(queryBuilder = new QueryBuilder()) {
-        queryBuilder.limit(1);
-        return this.model.findOne(queryBuilder.toImplementation());
+    async findOne(findQuery = {}) {
+        if (findQuery instanceof QueryBuilder) {
+            findQuery.limit(1);
+            findQuery = findQuery.toImplementation();
+        }
+        return this.model.findOne(findQuery);
     }
 
     /**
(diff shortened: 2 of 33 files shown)


lib/public/views/Logs/Create/LogCreationModel.js
```diff
@@ -32,6 +32,8 @@ export class LogCreationModel extends Observable {
 
         this._onCreation = onCreation;
 
+        this._logType = 'New log';
+        this._titleLabelText = 'Title*';
```
martinboulais, line 36: I think these two information do not belongs to the model, because they can be fully deduced from `this._parentLog`. They should be computed directly in the view without using intermediate variable.

lib/public/views/Logs/Create/logCreationComponent.js
```diff
@@ -30,7 +30,7 @@ export const logCreationComponent = (creationModel) => {
     return [
         data.isFailure() && data.payload.map(errorAlert),
         creationModel.parentLog.match({
-            NotAsked: () => errorAlert([{ title: 'Missing parent log' }]),
+            NotAsked: () => logCreationForm(creationModel),
             Loading: () => spinner(),
             Success: () => logCreationForm(creationModel),
```
martinboulais, line 35: Here, I would pass the parent log as a parameter, see later how to use it.

lib/public/views/Logs/Create/logCreationComponent.js
```diff
@@ -46,7 +46,7 @@ export const logCreationComponent = (creationModel) => {
  */
 const logCreationForm = (creationModel) => [
     h('.w-100.flex-row.mv2.justify-between', [
-        h('h3', 'New log'),
+        h('h3', `${creationModel.logType}`),
```
martinboulais, line 49: Supposing you have the parent log as a parameter, you can use here:
``${parentLog ? `Reply to log: ${parentLog.title}` : 'New log'}``

lib/public/views/Logs/Create/logCreationComponent.js
```diff
@@ -96,12 +96,13 @@ const labeledInput = (labelAttributes, labelText, content, containerAttributes)
  */
 const titlePanel = (logCreationModel) => labeledInput(
     { for: 'title' },
-    'Title*',
+    `${logCreationModel.titleLableText}`,
```
martinboulais, line 99: Same here: 
`${parentLog ? 'Reply title' : 'Title'}`

lib/public/views/Logs/Create/LogCreationModel.js
```diff
@@ -51,7 +53,7 @@ export class LogCreationModel extends Observable {
                 },
             );
         } else {
-            this._parentLog = RemoteData.success(null);
+            this._parentLog = RemoteData.notAsked();
```
martinboulais, line 56: I strongly disagree in the fact to use the `RemoteData.notAsked()` to define if there is a parent log or not. In my opinion we base the code on an implicit information (if the data has not been asked, it means that there is no reason to ask for it). But this relation may be at some point broken, for example if we delay the loading of the parent log, forget to update it and so on.
The `RemoteData` should only carry information about the **state of a remote information**, but nothing about the information itself, because else we are mixing two information for very different contexts in a variable (the state of the parent and the actual parent in the case it is null), and splitting another information in two variable (the actual parent log that may be null is split between the remote data and its payload)

martinboulais (changes requested):

PR #1035, opened 2023-06-28
[O2B-907] Add property tags to store color

GMvdDonck:
#### I have a JIRA ticket
- [x] PR labels are selected

Changes made to the database:
Added color field to tags


lib/database/migrations/20230628131012-add-tag-color.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+    // eslint-disable-next-line require-jsdoc
+    async up(queryInterface, Sequelize) {
+        return queryInterface.addColumn('tags', 'color', {
+            allowNull: true,
+            type: Sequelize.STRING,
```
martinboulais, line 9: I think that here you can also check that the string is 6 characters wide

lib/domain/dtos/CreateTagDto.js
```diff
@@ -32,7 +32,12 @@ const BodyDto = Joi.object({
         .optional()
         .allow('', null)
         .email({ multiple: true }),
-
+    color: Joi.string()
+        .min(6)
+        .max(6)
+        .trim()
+        .optional()
+        .allow('', null),
```
martinboulais, line 40: Here you should probably allow only null

lib/domain/dtos/GetAllTagsDto.js
```diff
@@ -19,6 +19,7 @@ const FilterDto = Joi.object({
     partialText: Joi.string().trim().optional(),
     emails: Joi.string().trim().optional(),
     mattermosts: Joi.string().trim().optional(),
+    colors: Joi.string().trim().optional(),
```
martinboulais, line 22: This is for filtering on tags page, I am not sure this is necessary

lib/domain/dtos/UpdateTagDto.js
```diff
@@ -27,6 +27,12 @@ const BodyDto = Joi.object({
         .optional()
         .allow('', null)
         .email({ multiple: true }),
+    color: Joi.string()
+        .min(6)
+        .max(6)
+        .trim()
+        .optional()
+        .allow('', null),
```
martinboulais, line 35: Same here I think that you should not allow ''

lib/public/views/Tags/Create/TagCreationModel.js
```diff
@@ -64,6 +64,7 @@ class TagCreationFormData extends Observable {
         this._description = '';
         this._email = '';
         this._mattermost = '';
```
martinboulais, line 66: This is frontend only and should probably not be part of this PR

martinboulais (changes requested):

PR #1056, opened 2023-07-10
[O2B-795] Store run definition in database

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

- The run definition is now stored in the database and updated when it makes sense

Changes made to the database:
- A column "definition" has been added to the runs table, and need to be updated manually when updating runs or related lhcFill or runType


lib/server/services/run/RunService.js
```diff
@@ -135,11 +137,23 @@ class RunService {
...
+            runType = await getOrCreateRunType({ name: runTypeName });
             newRun.runTypeId = runType.id;
         }
 
+        let lhcFill = null;
+        if (newRun.fillNumber) {
+            lhcFill = await getLhcFill(newRun.fillNumber);
+        }
+
+        newRun.definition = getRunDefinition({
```
graduta, line 151: I do not think that `createRun` is the correct place to calculate the runDefinition because AliECS on this step does not send yet the `timeO2Start` and `timeTrgStart` which means that the `getRunDefinition` will not be able to identify an overlap between the run and the stable beam.

This information comes only at a latter stage in `updateRun`, right?

graduta (changes requested):

lib/database/migrations/20230706083148-add-run-definition-column.js
```diff
@@ -0,0 +1,114 @@
...
+      AND readout_cfg_uri LIKE '%replay%'
+      AND (readout_cfg_uri LIKE '%pp%' OR readout_cfg_uri LIKE '%pbpb%')
+`;
+
+const SET_CALIBRATION_DEFINITION = `
+    UPDATE runs r
+        INNER JOIN run_types rt ON r.run_type_id = rt.id
+    SET r.definition = 'CALIBRATION'
+    WHERE r.definition IS NULL
+        AND rt.name LIKE 'CALIBRATION_%'
```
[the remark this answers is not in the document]
martinboulais: You're right, I updated it

lib/server/services/run/RunService.js
martinboulais: As discussed together, it's in the low level services:
In the LHC fill update and in the run creation and update

PR #1061, opened 2023-07-14
[O2B-927] Hide duplicated and verbose information in log tree view

joetorsney:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Removes log titles if they're the same as the root log
- Removes the panel on the left showing all data about the log
- Adds a string with the author and creation date of the log
- Adds tags, attachment, and run information if the log is expanded

- Added a component `clipboardTarget` which wraps other components with behavior to copy text to the clipboard.

Changes made to the database:
- None

Todo
- [x] Style tags as badges
- [x] Check that tests pass / fix those that don't
- [x] Add link button

### DIFF
--- a/lib/public/components/Log/linkButton.js
+++ b/lib/public/components/Log/linkButton.js
@@ -13,14 +13,19 @@
 
 import { h } from '/js/src/index.js';
 import {iconLinkIntact} from '/js/src/icons.js'
+import { clipboardTarget } from '/components/common/clipboardTarget.js';
 
 /**
  * Copies the url of the log to the clipboard
  * @param {onClick} the function to be called when clicked
  * @return {Component} The link button.
  */
-export const linkButton = (log, onClick) => h('a.btn.btn-primary', {
-    id: `copy-${log.id}`,
-    style: 'margin-right: 5px',
-    onclick: onClick,
-}, iconLinkIntact());
+export const linkButton = (log) => {
+    const url = new URL(window.location.href);
+    url.searchParams.set('id', log.id);
+
+    return clipboardTarget(url, h('div.btn.btn-primary', {
+        id: `copy-${log.id}`,
+        style: 'margin-right: 5px',
+    }, iconLinkIntact()));
+}
--- a/lib/public/components/Log/linkButton.js
+++ b/lib/public/components/Log/linkButton.js
@@ -12,12 +12,12 @@
  */
 
 import { h } from '/js/src/index.js';
-import {iconLinkIntact} from '/js/src/icons.js'
+import { iconLinkIntact } from '/js/src/icons.js';
 import { clipboardTarget } from '/components/common/clipboardTarget.js';
 
 /**
  * Copies the url of the log to the clipboard
- * @param {onClick} the function to be called when clicked
+ * @param {Log} log the log object that this button corresponds to
  * @return {Component} The link button.
  */
 export const linkButton = (log) => {
@@ -28,4 +28,4 @@ export const linkButton = (log) => {
         id: `copy-${log.id}`,
         style: 'margin-right: 5px',
     }, iconLinkIntact()));
-}
+};
--- a/lib/public/components/Log/linkButton.js
+++ b/lib/public/components/Log/linkButton.js
@@ -1,0 +1,26 @@
+/**
+ * @license
+ * Copyright CERN and copyright holders of ALICE O2. This software is
+ * distributed under the terms of the GNU General Public License v3 (GPL
+ * Version 3), copied verbatim in the file "COPYING".
+ *
+ * See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { h } from '/js/src/index.js';
+import {iconLinkIntact} from '/js/src/icons.js'
+
+/**
+ * Copies the url of the log to the clipboard
+ * @param {onClick} the function to be called when clicked
+ * @return {Component} The link button.
+ */
+export const linkButton = (log, onClick) => h('a.btn.btn-primary', {
+    id: `copy-${log.id}`,
+    style: 'margin-right: 5px',
+    onclick: onClick,
+}, iconLinkIntact());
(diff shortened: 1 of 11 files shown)


lib/public/components/Log/linkButton.js
```diff
@@ -0,0 +1,31 @@
...
+import { h } from '/js/src/index.js';
+import { iconLinkIntact } from '/js/src/icons.js';
+import { clipboardTarget } from '/components/common/clipboardTarget.js';
+
+/**
+ * Copies the url of the log to the clipboard
+ * @param {Log} log the log object that this button corresponds to
+ * @return {Component} The link button.
+ */
+export const linkButton = (log) => {
```
martinboulais, line 23: As I said, you should NEVER do supposition on the context in the components. 
When you will use this component to be able to copy log link from runs details page, log tab, will this still work?

Also, when I will want to create a run link button, can I still name it `linkButton`? It will become hard to chose when using autocomplete, and I will have to look at the import declaration when reading code using this!

lib/public/components/common/clipboardTarget.js
```diff
@@ -0,0 +1,27 @@
...
+
+import { h } from '/js/src/index.js';
+
+/**
+ * Copies a given text to the clipboard if the window is secure
+ * @param {copyText} copyText the text to copy
+ * @param {Component} children components
+ * @return {Component} the children components wrapped in the clipboardTarget
+ */
+export const clipboardTarget = (copyText, children) => h('a', { onclick: (e) => {
```
martinboulais, line 22: Why using an `a` here and not a simple button?
As a rule of thumb, if you do not have an `href` you should not have a link.

lib/public/components/Log/replyButton.js
```diff
@@ -25,7 +25,7 @@ export const replyButton = (log) => frontLink(
     { parentLogId: log.id },
     {
         id: `reply-to-${log.id}`,
+        style: 'margin-right: 5px',
```
martinboulais, line 28: (almost) never use inline styles, and absolutely never use hard-coded pixels dimensions: it prevent UI from having an homogenous display, it prevent anyone to properly zoom in and so on.
I know it's legacy code but you touched it so now you're cursed and you have to clean it.
You have utility classes such as `m1`, `m2`, `m3` to apply pre-defined margins.

lib/public/components/Log/logCollapsed.js
```diff
@@ -62,32 +83,35 @@ const textBox = ({ text, id }, onClick) =>
...
-                        key, {
-                            name,
-                            format,
-                            visible,
-                        },
-                    ]) => visible && h(`tr#log-${log.id}-${key}`, [
-                        h('td.text-right', { style: 'font-weight: bold' }, `${name}:`),
-                        h('td', format ? format(log[key]) : log[key]),
-                    ]))),
+    // eslint-disable-next-line require-jsdoc
```
martinboulais, line 93: I think this line can be removed

lib/public/components/Log/logCollapsed.js
```diff
@@ -62,32 +83,35 @@ const textBox = ({ text, id }, onClick) =>
...
+    return h(
+        `#log-${log.id}.w-100.m1.p3.shadow-level1${highlight ? '.b1.b-primary' : ''}`,
+        h(
+            'div.flex-wrap.justify-between.mv1',
+            showLogTitle ? h(`h4#log-${log.id}-title`, log.title) : '',
+            h('div.flex-row.justify-between.w-100', [
```
martinboulais, line 99: I'm pretty sure you can do without w-100 (this class should be avoided) using flex grow for example (I just see that it's in the framework example... which is bad)

lib/public/components/Log/logCollapsed.js
```diff
@@ -62,32 +83,35 @@ const textBox = ({ text, id }, onClick) =>
...
+                h('div.flex-column.log-details-collapsed', [
+                    h('em', `${logFields.author.format(log.author)}, (${logFields.createdAt.format(log.createdAt)})`),
+                    ['tags'].map((key) =>
+                        h(`div#log-${log.id}-${key}.flex-row`, [
+                            h('div.mr2', `${logFields[key].name}:`),
+                            logFields[key].format(log[key]),
+                        ])),
+                ]),
+                h('div.flex-row', { style: 'align-items: flex-start' }, [
```
martinboulais, line 108: Please no inline styles, you have a class to align items at start

lib/public/components/Log/logLinkButton.js
```diff
@@ -20,12 +20,10 @@ import { clipboardTarget } from '/components/common/clipboardTarget.js';
  * @param {Log} log the log object that this button corresponds to
  * @return {Component} The link button.
  */
-export const linkButton = (log) => {
+export const logLinkButton = (log) => {
+    // TODO: absolute url
```
martinboulais, line 24: Please do this todo in this PR
You can extract a function `getFrontUrl` from `frontLink` component to do so
Feel free to ask me if you need any help

lib/public/components/common/clipboardTarget.js
```diff
@@ -14,12 +14,13 @@
...
 /**
- * Copies a given text to the clipboard if the window is secure
+ * Copies a given text to the clipboard if the window is secure. Wraps child components.
  * @param {copyText} copyText the text to copy
+ * @param {string} parentSelector the css selector of the parent node e.g. "button.btn"
  * @param {Component} children components
  * @return {Component} the children components wrapped in the clipboardTarget
  */
-export const clipboardTarget = (copyText, children) => h('a', { onclick: (e) => {
+export const clipboardTarget = (copyText, parentSelector, children) => h(parentSelector, { onclick: (e) => {
```
martinboulais, line 23: Here I think we should not mix the behavior-based component (copy to clipboard) from actual dom. Plus, taking a mithril vnode description as argument makes the code completely dependent on the mithril syntax, and blocks us if for example at some point we use jsx.
Could you please try to remove the 'parentSelector' and simply wrap children in a div?
The `button.btn` should simply be included in the children.

martinboulais (changes requested):

lib/public/components/Log/logLinkButton.js
joetorsney: Ah of course, thanks for catching that!

martinboulais (approved):

PR #1072, opened 2023-07-24
[O2B-929] Run details should list all detectors

emilyslade123:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Run details page shows all detectors as grey (not present) or green/red (good/bad quality)


lib/public/views/Runs/format/formatRunDetectors.js
```diff
@@ -29,23 +29,51 @@ export const formatRunDetectors = (run, detailsModel) => {
...
+        Loading: () => [],
+        Success: (data) => data,
+        Failure: (error) => error,
+    });
+
+    const _allAndSelectedDetectorTypes = _detectorTypes.map(({ id, name }) => {
+        const isPresent = detectorsQualities.map((obj) => obj.id).includes(id);
+        const quality = isPresent ? detectorsQualities.filter((obj) => obj.id == id)[0].quality : null;
+        return { id, name, quality, isPresent };
+    });
```
martinboulais, line 43: Be careful here: `_detectortypes` can be a list of errors.
What you should do instead is to wrap the whole component in the match `Success` case, because it does not makes sense in any other case.

lib/public/views/Runs/format/formatRunDetectors.js
```diff
@@ -29,23 +29,51 @@ export const formatRunDetectors = (run, detailsModel) => {
         return '-';
     }
 
+    const _detectorTypes = detailsModel.detectorTypes.match({
```
martinboulais, line 32: The convention of `_` prefix is for classes private members. Here you are not in a class, so you should simply name the variable `detectorTypes`.
However, I would suggest to name it `allDetectors` because it's actually a list of detectors, there is no such thing as detector type (please adapt the model accordingly).

lib/public/views/Runs/format/formatRunDetectors.js
```diff
@@ -29,23 +29,51 @@ export const formatRunDetectors = (run, detailsModel) => {
...
+    const _allAndSelectedDetectorTypes = _detectorTypes.map(({ id, name }) => {
+        const isPresent = detectorsQualities.map((obj) => obj.id).includes(id);
+        const quality = isPresent ? detectorsQualities.filter((obj) => obj.id == id)[0].quality : null;
+        return { id, name, quality, isPresent };
+    });
+
     const detectorsBadges = h(
         '.g2.flex-row.flex-grow.flex-wrap.justify-end',
-        detectorsQualities.map(({ id, name, quality }) => {
+        _allAndSelectedDetectorTypes.map(({ id, name, quality, isPresent }) => {
```
martinboulais, line 47: This code works fine, but I would suggest another implementation. The objective is to loop only once over the list of all detectors, and once over the list of detectors actually linked to the run:

Instead of the actual `_allAndSelectedDetectorTypes`, I would create a map indexing the detector quality (value) by the detector id (name).
Then, here, I would loop over all the detectors, and get the quality from the map (if it's undefined, it means that the detector is not included)

lib/public/views/Runs/format/formatRunDetectors.js
```diff
@@ -29,23 +29,51 @@ export const formatRunDetectors = (run, detailsModel) => {
...
+                    icon = iconX();
+                    break;
+                default:
+                    color = 'gray';
+                    icon = iconBan();
+                    break;
+            }
             return h(
-                `span.detector-badge.flex-row.justify-center.lh1.g1.br2.pv1.ph2.b1.b-${color}.${color}`,
+                `span.detector-badge.flex-row.justify-center.lh1.g1.br2.pv1.ph2.b1.b-${color}.${color}${isPresent ? '.isPresent' : ''}`,
```
martinboulais, line 68: I think that you should not create classes specifically for tests (it's a mistake commonly done in bookkeeping code), test should never impact the actual code.
Instead in the tests, you know from the seeders what detectors and qualities to expect, you can then check the color that is applied.

martinboulais (changes requested):

lib/public/views/Runs/format/formatRunDetectors.js
```diff
@@ -29,85 +29,81 @@ export const formatRunDetectors = (run, detailsModel) => {
...
-                    '.scroll-y',
-                    detectorsQualities.map(({ id, name, quality }) => h(
-                        '.flex-row.justify-between.g4.p2',
-                        [
-                            name,
-                            h(
-                                '.flex-row.g2.items-center',
-                                RUN_DETECTOR_QUALITIES.map((qualityOption) => h('label.form-check-label.flex-row.g2', [
+            if (detailsModel.isEditModeEnabled && (run.timeTrgEnd || run.timeO2End)) {
+                detectorsBadges.children = detectorsBadges.children.filter((elem) => !elem.attrs.className.includes('b-gray'));
```
martinboulais, line 68: This is a nice workaround, but still: here you have a component directive that depend on a css class. This should never happen: component description should result in dom description, never the other way around.
Here you need to think from a higher point of view, and maybe split the component with the `isEditModeEnabled` earlier in two separated component
We can discuss about it if you want

martinboulais (changes requested):

lib/public/views/Runs/format/formatRunDetectors.js
```diff
@@ -29,48 +29,56 @@ export const formatRunDetectors = (run, detailsModel) => {
         return '-';
     }
 
+    /**
+     * Creates a component containing the detector badges
+     *
+     * @param {List<Detector>} detectors list of detectors
```
martinboulais, line 35: The syntax for arrray is `Detector[]` (prefered) or `Array<Detector>`
However the type you pass is not a `Detector` array, but an array of detector qualities. Could you please create a typedef file in `lib/domain/entities`, and add in it the typedef:
```js
/**
 * @typedef RunDetectorQuality
 * @property {number} id
 * @property {string} name
 * @property {string[]} quality
 */
```

Then you can use it to type this param. (Please also use this type in `lib/domain/entities/Run` to replace line 67 property type)

martinboulais (changes requested):

martinboulais (approved):

PR #1078, opened 2023-07-27
O2B-937 About page enhancements

joetorsney:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Enhances the about page to include information and status of services including the Bookkeeping gui and the database.

- None

Changes made to the database:
- None

### DIFF
--- a/lib/public/Model.js
+++ b/lib/public/Model.js
@@ -26,7 +26,7 @@ import { debounce, INPUT_DEBOUNCE_TIME } from './utilities/debounce.js';
 import { userPreferencesStore } from './utilities/userPreferencesStore.js';
 import { registerFrontLinkListener } from './components/common/navigation/frontLinkListener.js';
 import { EosReportModel } from './views/EosReport/EosReportModel.js';
-import AboutModel from './views/About/About.js'
+import AboutModel from './views/About/About.js';
 
 /**
  * Root of model tree
@@ -118,8 +118,8 @@ export default class Model extends Observable {
         /**
          * @type {AboutModel}
          */
-        this.AboutModel = new AboutModel()
-        this.AboutModel.bubbleTo(this)
+        this.AboutModel = new AboutModel();
+        this.AboutModel.bubbleTo(this);
 
         // Setup router
         this.router = new QueryRouter();
@@ -227,6 +227,7 @@ export default class Model extends Observable {
                 break;
             case 'about-overview':
                 this.AboutModel.fetchAppInfo();
+                this.AboutModel.fetchDbInfo();
                 break;
             case 'eos-report-create':
                 this.eosReportModel.loadCreation(this.router.params.shiftType);
--- a/lib/public/Model.js
+++ b/lib/public/Model.js
@@ -26,6 +26,7 @@ import { debounce, INPUT_DEBOUNCE_TIME } from './utilities/debounce.js';
 import { userPreferencesStore } from './utilities/userPreferencesStore.js';
 import { registerFrontLinkListener } from './components/common/navigation/frontLinkListener.js';
 import { EosReportModel } from './views/EosReport/EosReportModel.js';
+import AboutModel from './views/About/About.js'
 
 /**
  * Root of model tree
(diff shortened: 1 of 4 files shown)


lib/public/views/About/About.js
```diff
@@ -11,19 +11,98 @@
...
  */
-export default class Overview extends Observable {
+export default class AboutModel extends Observable {
     /**
-     * The constructor of the Overview model object
-     * @param {Object} model Pass the model to access the defined functions
-     * @returns {Object} Constructs the Overview model
+     * The constructor for the About model object
+     * @param {Object} model Pass the parent model to access the defined functions
+     * @return {Object} Constructs the About model
```
graduta, line 24: The JSDOC recognizes `returns` as the official tag

lib/public/views/About/About.js
```diff
@@ -11,19 +11,98 @@
...
-import { Observable } from '/js/src/index.js';
+import { getRemoteData } from '../../utilities/fetch/getRemoteData.js';
+import { Observable, RemoteData } from '/js/src/index.js';
 
 /**
- * Model representing handlers for homePage.js
+ * Model representing handlers for aboutPage.js
  */
-export default class Overview extends Observable {
+export default class AboutModel extends Observable {
```
graduta, line 20: While this was the old convention of the project, the current convention is that we use named exports rather than default exports.
See `EosReportModel` on how to adapt

lib/public/Model.js
```diff
@@ -114,6 +115,12 @@ export default class Model extends Observable {
         this.eosReportModel = new EosReportModel(this);
         this.eosReportModel.bubbleTo(this);
 
+        /**
+         * @type {AboutModel}
+         */
+        this.AboutModel = new AboutModel();
```
graduta, line 121: For variables we use `camelCase`, thus variables start with lower case

graduta (changes requested):
Currently, if the request fails then the error message is not as clear as with the success request.
I would recommend you add the `ok` badge failed as well or something similar for the failed request.
Referring to usecase below
image

lib/public/views/About/About.js
```diff
@@ -11,19 +11,88 @@
...
+        }
+        return service;
+    }
+
+    /**
+     * Set the app info and notify
+     * @param {RemoteData} appInfo information about the app
+     * @returns {void}
+     */
+    set appInfo(appInfo) {
```
graduta, line 68: As the setter is not being used at all, I would recommend to remove it. 
There is no point in allowing modification if we do not plan to do it

lib/public/views/About/About.js
```diff
@@ -11,19 +11,88 @@
...
+    }
+
+    /**
+     * Fetches information about the service determined by the api path
+     * @param {string} apiPath the path to the endpoint for this service's status
+     * @returns {Object} an object which contains the version (if the request was successful)
+     *                   and the status of the service
+     */
+    async fetchServiceInfo(apiPath) {
+        let service = RemoteData.loading();
```
graduta, line 50: This method should also receive by reference the serviceInfo object so that the changes from `loading` to `success` or `error` are visible in the UI.
Currently you create a variable within this scope only and also, you do not notify the model on the variable change.

lib/public/views/About/About.js
```diff
@@ -11,19 +11,88 @@
...
+    get appInfo() {
+        return this._appInfo;
+    }
+
+    /**
+     * Set the database info and notify
+     * @param {RemoteData} dbInfo information about the database
+     * @returns {void}
+     */
+    set dbInfo(dbInfo) {
```
graduta, line 86: Same for this `setter`. IF we are not using it, let's not add dead code

lib/public/views/About/Overview/index.js
```diff
@@ -11,28 +11,110 @@
...
+};
 
-    return h('.flex-row.pv2.w-100', table(data, aboutColumns));
+/**
+ * Returns a view of information panels for each service
+ * @param {object} model Pass the model to access the defined functions
+ * @returns {vnode} the view of services
+ */
+const aboutOverview = (model) => {
+    const { appInfo } = model.aboutModel;
```
graduta, line 115: You can extract both `appInfo` and `dbInfo` in the same statement.
`const {appInfo, dbInfo} = model.aboutModel`
or 
`const {aboutModel: {appInfo, dbInfo}} = model`

graduta (changes requested):
Changes are in the right direction, good job!
With a few extra improvements, you will be able to provide a better experience for the users but also for developers

graduta (approved):

PR #1081, opened 2023-08-01
[O2B-944] Log details page style improvements

joetorsney:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Further improves the style of the log details page, following from https://github.com/AliceO2Group/Bookkeeping/pull/1061
- Adds feedback to the user when the log link is successfully copied

- Adds a primary button component to the common component library
- Adds more tests on the log details page
- Refactored the log component to share common code between expanded and collapsed logs.

Changes made to the database:
- None


lib/public/views/Logs/Details/LogTreeViewModel.js
```diff
@@ -72,6 +72,31 @@ export class LogTreeViewModel extends Observable {
...
 
+    /**
+     * Given a log id and a value, sets the showCopyUrlSuccessContent indicator for that log,
+     * which determines whether to show feedback to the user that the log url was copied to the clipboard
+     * @param {number} id the id of the log to set the indicator for
+     * @param {boolean} value the value of the indicator
+     * @return {void}
+     */
+    setShowCopyUrlSuccessContent(id, value) {
+        this._logTreeLeafs.payload.find((leaf) => leaf.log.id == id).showCopyUrlSuccessContent = value;
```
graduta, line 83: Please use `===` the strictly equality operator

lib/public/views/Logs/Details/LogTreeViewModel.js
```diff
@@ -118,7 +143,7 @@ export class LogTreeViewModel extends Observable {
      * Retrieve the tree of a given log from the API
      *
      * @param {number} selectedLogId The id of the log for which tree must be retrieved
-     * @returns {undefined} Injects the data object with the response data
+     * @return {void} Injects the data object with the response data
```
graduta, line 146: JSDOC recommendation is to use `returns` rather than `return`
https://jsdoc.app/tags-returns.html 
Please update accordingly on the other changes you have made

lib/public/views/Logs/Details/LogTreeViewModel.js
```diff
@@ -137,7 +162,7 @@ export class LogTreeViewModel extends Observable {
              */
             const parseTreeNode = (node, level) => {
                 const { children, /** @type {Log} */ ...log } = node;
-                logTreeLeafs.push({ log, level });
+                logTreeLeafs.push({ log: log, level: level, showCopyUrlSuccessContent: false });
```
graduta, line 165: When the key is the same as the variable containing the value, there is no need to specify it. Thus, for `log` and `level` it is ok to stay as `{ log, level, showCopyUrlSuccessContent: false }`

lib/public/components/Log/log.js
```diff
@@ -0,0 +1,160 @@
...
+);
+
+/**
+ * Returns a collapse button.
+ *
+ * @param {Log} log the log to collapse
+ * @param {function} onClick function called when button is activated
+ * @return {Component} the collapse button component
+ */
+const collapseButton = (log, onClick) => h('a.btn.btn-primary', {
```
graduta, line 89: There is no need for this button to be a hyperlink and it can be a normal HTML `button

lib/public/components/Log/log.js
```diff
@@ -0,0 +1,160 @@
...
+    ]));
+
+/**
+ * Returns a log's content display
+ *
+ * @param {Log} log the log to display
+ * @param {Boolean} isCollapsed an indicator as to whether the log is collapsed
+ * @return {Component} the log's content display
+ */
+const textBox = (log, isCollapsed) => h(
```
graduta, line 72: This component should be extracted in an individual component and should have improved documentation on what the content of the log includes
For example, it is not clear why we have `logContent` and `logDetails` with both being better suited as individual components for future reusesability

graduta (changes requested):
Very good improvements for the readability page. With a few changes it will be ready to go

graduta (approved):

PR #1087, opened 2023-08-07
[O2B-942] Add front end environment log filtering

emilyslade123:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Can filter logs by environments

### DIFF
--- a/lib/public/views/Logs/Logs.js
+++ b/lib/public/views/Logs/Logs.js
@@ -171,6 +171,10 @@ export default class LogModel extends Observable {
         this.runFilterValues = [];
         this._runFilterRawValue = '';
 
+        this.environmentFilterOperation = 'AND';
+        this.environmentFilterValues = [];
+        this._environmentFilterRawValue = '';
+
         this._pagination.reset();
 
         if (fetch) {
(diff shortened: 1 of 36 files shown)


lib/public/views/Logs/Logs.js
```diff
@@ -224,6 +229,38 @@ export default class LogModel extends Observable {
...
+    }
+
+    /**
+     * Sets the environment filter
+     * @param {String} rawEnvironments The environments to apply to the filter
+     * @returns {undefined}
+     */
+    setEnvFilter(rawEnvironments) {
+        this._environmentFilterRawValue = rawEnvironments;
+        const environmentRegex = /([0-9a-zA-Z]+(?=,?))/g; // Matches comma-separated alphanumeric strings
```
martinboulais, line 255: Maybe you should also allow for spaces around the env ids?

martinboulais (changes requested):

lib/public/views/Logs/Logs.js
martinboulais: Yes I was specifically thinking between the env id and the comma. This should of course never happen, but sometimes copy-paste may add space and this can be frustrating. 
`envId1 , envId2` should work as `envid1,envId2`

martinboulais (approved):

PR #1093, opened 2023-08-08
[O2B-945] Removed active filters bar on Log and Runs page

emilyslade123:
#### I have a JIRA ticket
- [ ] PR labels are selected

Notable changes for users:
- Removed Active Filters bar on Run Overview page and Log Overview page


graduta (approved):

graduta (changes requested):
I believe you should remove also code within the scope of this view component.
For example, `activeFilters` array is used only here, the same for the getter for it `getActiveFilters()`

This code is not used at this point, thus it should be removed as well

martinboulais (approved):

PR #1095, opened 2023-08-11
[O2B-918] Autofill EoS report

emilyslade123:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- End of Shift reports are pre-filled with information from the previous shifter


lib/server/services/shift/ShiftService.js
```diff
@@ -64,6 +65,29 @@ class ShiftService {
             typeSpecific: null,
         };
 
+        // Find all EoS reports from the last 16 hours and sort by most recent first
+        const pastShift = { start: new Date() - SHIFT_DURATION * 2, end: new Date() };
```
martinboulais, line 69: This is not correct, because here you will have shift that start exactly 16 hours ago, which is not the reality. If you run this code at 2PM you want `start` to be 23PM the day before, not 22PM for example.

Instead I would suggest to simply use the current shift, something like this should work:

```js
const pastShift = getShiftFromTimestamp(shift.start - SHIFT_DURATION);
```

You should extract this in a function such as `getPreviousShift` and add test to it, checking corner cases such as exact 3PM

lib/server/services/shift/ShiftService.js
```diff
@@ -64,6 +65,29 @@ class ShiftService {
...
+                new Date(b.createdAt) - new Date(a.createdAt);
+            });
+
+        // Filter for the correct tags and EoS title and select the most recent report
+        const reportFormat = `End of shift report - ${shiftType} - `;
+        const tags = getEosReportTagsByType(shiftType);
+        const allEosReports = prevShiftIssues.filter((report) =>
+            tags.every((tag) => report.tags.map((x) => x.dataValues.text).includes(tag))
+            && report.title.startsWith(reportFormat));
+        const eosReport = allEosReports.length > 0 ? allEosReports[0] : { text: '' };
```
martinboulais, line 82: As we discussed, here I would rather :
- Extract a function to format EoS report title : https://github.com/AliceO2Group/Bookkeeping/blob/main/lib/server/services/eosReport/formatEosReport.js#L34
- Use it to get the expected title of the previous EoS report

lib/server/services/shift/ShiftService.js
```diff
@@ -64,6 +65,29 @@ class ShiftService {
...
+
+        // Extract the 'For next shifter' information
+        const infoFromPreviousShifterRegex = /(?<=\n\n### For next shifter\n)([\S\s]*)(?=\n\n###)/g;
```
martinboulais, line 85: I think this can be simplified: `/### From previous shifter\n([\s\S]*)\n\n### For next shifter/`

lib/public/views/EosReport/create/eosReportCreationComponent.js
```diff
@@ -102,8 +102,11 @@ const eosReportCreationForm = (reportType, formData, shiftData) => [
         }, () => null)(),
         h('#from-previous-shifter.panel', [
             h('label.panel-header', 'Info from previous shifter'),
+            !formData.infoFromPreviousShifter && shiftData.previousInfo
```
martinboulais, line 105: please use more accurate variables name, such as `shiftData.infoFromPreviousShifter`

test/mocks/base-mock-eos-report.js
```diff
@@ -69,6 +72,65 @@ exports.genericCustomizedEosReportRequest = {
     infoForRmRc: genericCustomizedEosReport.infoForRmRc,
 };
 
+// eslint-disable-next-line require-jsdoc
+const genericCustomTimeEosReport = (time, information) => ({
```
martinboulais, line 76: What is the purpose of these mocks?

martinboulais (changes requested):

lib/server/services/shift/ShiftService.js
emilyslade123: It needs the positive lookahead and lookbehind so they're not included in the capture groups

lib/server/services/shift/ShiftService.js
```diff
@@ -53,40 +54,48 @@ class ShiftService {
...
-            ? [ret.previousInfo] = infoFromPreviousShifterAllMatches
-            : ret.previousInfo = '';
+        // Find EoS reports from the last shift with the correct title
+        const timestamp = shiftStart || shift.start;
+        const pastShift = getShiftFromTimestamp(timestamp - SHIFT_DURATION);
+        const title = formatEosReportTitle(pastShift, shiftType);
+
+        const prevShiftIssues = await getShiftIssuesByTitle(pastShift, title);
+
+        if (prevShiftIssues.length !== 1) {
```
martinboulais, line 78: You have two cases here: either you have no EoS report, either you have more than one

test/mocks/base-mock-eos-report.js
emilyslade123: To be able to properly test the autofill feature from 8 hours ago

test/lib/server/services/eosReport/EosReportService.test.js
```diff
@@ -109,15 +109,21 @@ module.exports = () => {
...
-        const response = await shiftService.getShiftData({ userId: 1 }, ShiftTypes.ECS);
-        expect(response.previousInfo).to.equal(info);
+        const response = await shiftService.getShiftData({ userId: 1 }, ShiftTypes.ECS, future);
+        expect(response.previousInfo.value).to.equal(info);
+    });
+
+    it ('should throw an error in previousInfo if no previous EoS report is found', async () => {
+        const response = await shiftService.getShiftData({ userId: 1 }, ShiftTypes.SLIMOS, Date.now());
+        expect(response.previousInfo.errorMessage).to.equal('No SLIMOS EoS reports found from the previous shift');
     });
```
martinboulais, line 127: You are changing shift service, so your tests must be about ShiftService. 
You do not need any customEoS report object, you only need to:
- create a log 8 hours ago with a fake EoS report (containing only the correct title and a body that match your regex)
- call your UserService function 

Do the same but on the frontend side

test/lib/server/services/eosReport/EosReportService.test.js
```diff
@@ -109,15 +109,21 @@ module.exports = () => {
...
 
-        const response = await shiftService.getShiftData({ userId: 1 }, ShiftTypes.ECS);
-        expect(response.previousInfo).to.equal(info);
+        const response = await shiftService.getShiftData({ userId: 1 }, ShiftTypes.ECS, future);
+        expect(response.previousInfo.value).to.equal(info);
+    });
+
+    it ('should throw an error in previousInfo if no previous EoS report is found', async () => {
+        const response = await shiftService.getShiftData({ userId: 1 }, ShiftTypes.SLIMOS, Date.now());
+        expect(response.previousInfo.errorMessage).to.equal('No SLIMOS EoS reports found from the previous shift');
```
martinboulais, line 126: Please add a test for all the cases you handle (0 EoS report, 2 EoS report, invalid body and so on)

martinboulais (changes requested):

test/lib/server/services/eosReport/EosReportService.test.js
emilyslade123: Added extra test cases

test/lib/server/services/eosReport/EosReportService.test.js
emilyslade123: Simplified ShiftService tests

lib/server/services/eosReport/getShiftIssuesByTitle.js
```diff
@@ -0,0 +1,21 @@
...
+ */
+const { getLogsByTitle } = require('../log/getLogsByTitle.js');
+
+/**
+ * Returns the issues related to a given shift for a given tagsFilter based on shift type
+ *
+ * @param {String} title the title used to filter issues
+ * @return {Promise<SequelizeLog[]>} resolves with the found logs
+ */
+exports.getShiftIssuesByTitle = async (title) => getLogsByTitle(title, { rootOnly: true });
```
martinboulais, line 21: I think this fonction is not necessary anymore, please directly use `getLogsByTitle`

lib/server/services/shift/ShiftService.js
```diff
@@ -58,40 +58,43 @@ class ShiftService {
...
         };
 
         // Find EoS reports from the last shift with the correct title
-        const timestamp = shiftStart || shift.start;
-        const pastShift = getShiftFromTimestamp(timestamp - SHIFT_DURATION);
+        const pastShift = getShiftFromTimestamp(shift.start - SHIFT_DURATION);
         const title = formatEosReportTitle(pastShift, shiftType);
 
-        const prevShiftIssues = await getShiftIssuesByTitle(pastShift, title);
+        const prevShiftIssues = await getShiftIssuesByTitle(title);
```
martinboulais, line 75: Please name variables in a more informative way. I think title should be `expectedPreviousReportTitle` and previousShiftIssues should be `previousShiftReports`

lib/server/services/shift/ShiftService.js
martinboulais: You should use the second value of match result, which is the actually matched items. Lookarounds makes the regex quite hard to read in my opinion, and I love regex...

lib/server/services/shift/ShiftService.js
```diff
@@ -58,40 +58,43 @@ class ShiftService {
...
+        } else if (prevShiftIssues.length > 1) {
+            ret.infoFromPreviousShifter.errorMessage = `Multiple ${shiftType} EoS reports found from the previous shift`;
         } else {
             const eosReportText = prevShiftIssues[0].text;
             if (!eosReportText) {
-                ret.previousInfo.errorMessage = 'EoS report from the previous shift is missing the text field';
+                ret.infoFromPreviousShifter.errorMessage = 'EoS report from the previous shift is missing the text field';
             } else {
-                const infoFromPreviousShifterRegex = /(?<=\n\n### For next shifter\n)([\S\s]*)(?=\n\n###)/g;
+                const infoFromPreviousShifterRegex = /(?<=\n\n### For next shifter\n)([\S\s]*?)(?=\n\n###)/g;
```
martinboulais, line 86: `*` is already allows for `0` match, I think `?` is not necessary here

test/lib/server/services/eosReport/EosReportService.test.js
```diff
@@ -326,4 +308,86 @@ module.exports = () => {
...
+        };
+
+        await eosReportService.createLogEntry(ShiftTypes.SLIMOS, request1, { userId: 1 });
+        await eosReportService.createLogEntry(ShiftTypes.SLIMOS, request2, { userId: 1 });
+
+        const response = await shiftService.getShiftData({ userId: 1 }, ShiftTypes.SLIMOS);
+        expect(response.infoFromPreviousShifter.errorMessage).to.equal('Multiple SLIMOS EoS reports found from the previous shift');
+    });
+
+    it ('should create a new EoS report autofilled with previous EoS report', async () => {
```
martinboulais, line 377: Could you update the test label please?

martinboulais (changes requested):

lib/server/services/shift/ShiftService.js
emilyslade123: The ? makes it a lazy operator so that I can test for duplicate fields (the tests don't work without it)

lib/server/services/shift/ShiftService.js
```diff
@@ -62,8 +65,41 @@ class ShiftService {
...
+            ret.infoFromPreviousShifter.errorMessage = `No ${shiftType} EoS reports found from the previous shift`;
+        } else if (previousShiftReports.length > 1) {
+            ret.infoFromPreviousShifter.errorMessage = `Multiple ${shiftType} EoS reports found from the previous shift`;
+        } else {
+            const eosReportText = previousShiftReports[0].text;
+            if (!eosReportText) {
+                ret.infoFromPreviousShifter.errorMessage = 'EoS report from the previous shift is missing the text field';
+            } else {
+                const infoFromPreviousShifterRegex = /\n\n### For next shifter\n([\S\s]*?)\n\n###/g;
+                const infoFromPreviousShifterAllMatches = [...eosReportText.matchAll(infoFromPreviousShifterRegex)].map((match) => match[1]);
```
martinboulais, line 87: const infoFromPreviousShifterAllMatches = [...eosReportText.matchAll(infoFromPreviousShifterRegex)].map((match) => match[1]);
Would probably throw if match is empty. You can use one more variable that would make it more clear:

const infoFromPreviousShifterMatch = eosReportText.match(/\n\n### For next shifter\n([\S\s]*?)\n\n###/);
if (infoFromPreviousShifterMatch) {
    const [_, infoFromPreviousShifter] = infoFromPreviousShifterMatch;
}

lib/server/services/shift/ShiftService.js
```diff
@@ -62,8 +65,41 @@ class ShiftService {
...
+        if (previousShiftReports.length === 0) {
+            ret.infoFromPreviousShifter.errorMessage = `No ${shiftType} EoS reports found from the previous shift`;
+        } else if (previousShiftReports.length > 1) {
+            ret.infoFromPreviousShifter.errorMessage = `Multiple ${shiftType} EoS reports found from the previous shift`;
+        } else {
+            const eosReportText = previousShiftReports[0].text;
+            if (!eosReportText) {
+                ret.infoFromPreviousShifter.errorMessage = 'EoS report from the previous shift is missing the text field';
+            } else {
+                const infoFromPreviousShifterRegex = /\n\n### For next shifter\n([\S\s]*?)\n\n###/g;
```
martinboulais, line 86: I think you can ignore if there is more than one match, we can simply take the first one (it should not happen, and it's not critical at all if we simply ignore the supplementary ones).

lib/server/services/log/getLogsByTitle.js
```diff
@@ -0,0 +1,41 @@
...
+ *  In applying this license CERN does not waive the privileges and immunities
+ *  granted to it by virtue of its status as an Intergovernmental Organization
+ *  or submit itself to any jurisdiction.
+ */
+
+const { LogRepository } = require('../../../database/repositories/index.js');
+const { Op } = require('sequelize');
+
+/**
+ * Returns the logs created in the given period that have at least one of the given tags
```
martinboulais, line 18: Please update the comment here

martinboulais (dismissed):

martinboulais (approved):

PR #1094, opened 2023-08-09
[O2B-946] Add Logs-LhcFills table to database

joetorsney:
#### I have a JIRA ticket
- [x] branch and/or PR name(s) include(s) JIRA ID
- [ ] issue has "Fix version" assigned
- [x] issue "Status" is set to "In review"
- [x] PR labels are selected

Notable changes for users:
- None

Notable changes for developers:
- Adds adapter for lhc fills to logs

Changes made to the database:
- Adds a table, migration, and seeders for the `log_lhcFills` table, enabling a many-to-many relationship between logs and lhc fills.

PR #1096, opened 2023-08-14
[O2B-947] Add API routes to create and retrieve logs with lhcFills

joetorsney:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- None

- Building from https://github.com/AliceO2Group/Bookkeeping/pull/1094, adds API routes to create and retrieve logs that have associated LHC fills

Changes made to the database:
- None


lib/server/services/lhcFill/getAllLhcFillsOrFail.js
```diff
@@ -0,0 +1,42 @@
...
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+const { getLhcFillByNumberOrFail } = require('./getLhcFillByNumberOrFail.js');
+
+/**
+ * Returns the list of lhcFill ids extracted from a comma separated (CSV) list of lhcFill numbers
+ *
+ * @param {number[]} [lhcFills] the CSV list of lhcFill numbers
+ * @return {Promise<null|number[]>} the list of lhcFill ids
```
martinboulais, line 19: When you return an array you should avoid returning nullable array. This may lead to weird bugs (especially when using js without proper IDE hints), because we may forgot to check for null.

lib/server/services/lhcFill/getAllLhcFillsOrFail.js
```diff
@@ -0,0 +1,42 @@
...
+ * @private
+ */
+exports.getLhcFillNumbersOrFail = async (lhcFills) => {
+    if (!lhcFills) {
+        return null;
+    }
+
+    // Error out if any value is NaN
+    if (lhcFills.some(Number.isNaN)) {
```
martinboulais, line 28: I think this kind of data validation should not go into low-level services, at this point you should be able to expect normal numbers

On a side note, be careful with `Number.isNaN`: https://stackoverflow.com/questions/33164725/confusion-between-isnan-and-number-isnan-in-javascript

lib/server/services/lhcFill/getAllLhcFillsOrFail.js
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Check if, for each lhcFill number, there is a lhcFill present in the DB with that lhcFill number
+     * and return those that are present.
+     */
+    const retrievedLhcFills = [];
+    for (const possibleFillNumber of lhcFills) {
+        const { fillNumber } = await getLhcFillByNumberOrFail(possibleFillNumber);
+        retrievedLhcFills.push(fillNumber);
+    }
+    return retrievedLhcFills;
```
martinboulais, line 41: If you fail on missing fill, you will be anyway returning the given LHC fill numbers. So either you return only the found one, either you throw, but not both.

lib/usecases/log/CreateLogUseCase.js
```diff
@@ -35,10 +36,16 @@ class CreateLogUseCase {
                 .map((x) => parseInt(x, 10))),
         ] : [];
 
+        const lhcFills = serializedLhcFills ? [
+            ...new Set(serializedLhcFills.split(',')
+                .map((x) => x.trim())
+                .map((x) => parseInt(x, 10))),
```
martinboulais, line 42: The filtering of NaN should be here I think

martinboulais (changes requested):

lib/server/services/lhcFill/getAllLhcFillsOrFail.js
joetorsney: I thought with this that the documentation was a bit misleading. `getLhcFillByNumberOrFail` does throw if one isn't found, and therefore `retrievedLhcFills` won't be returned!

PR #1102, opened 2023-08-18
[O2B-958] Display statistics as graph

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- Statistics page now displays efficiency statistics as a graph

Notable change for developpers: 
- A library has been created to render line-charts based on d3


graduta (approved):

graduta (changes requested):
Very nicely done!
There are a few leftovers which I think it would be good to tackle before proceeding:
- P as a type I believe it is a leftover
- we should display a message that there are no data points for a certain time range rather than an empty plot without any axes

PR #1110, opened 2023-08-23
[O2B-964] Fix for O2B-929 - show null run detector quality

emilyslade123:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Where runs have detectors without qualities, they are shown as being a different colour to detectors that weren't included in the run


lib/domain/enums/RunDetectorQualities.js
```diff
@@ -15,6 +15,7 @@
 const RunDetectorQualities = {
     GOOD: 'good',
     BAD: 'bad',
+    NONE: 'none',
```
martinboulais, line 18: This enum must reflect what is in the database, and if you add this value it does not. I think you should instead handle the case of null value directly.

martinboulais (changes requested):

lib/public/views/Runs/format/formatRunDetectors.js
```diff
@@ -40,7 +40,7 @@ export const formatRunDetectors = (run, detailsModel) => {
             '.g2.flex-row.flex-grow.flex-wrap.justify-end',
             detectors.map(({ id, name }) => {
                 const quality = detectorsQualities.find((x) => x.id == id)
-                    ? detectorsQualities.find((x) => x.id == id).quality
+                    ? detectorsQualities.find((x) => x.id == id).quality || 'none'
```
martinboulais, line 43: Here I would use an if, to make the code clearer because we start to combine a bit too much conditionals.

Here is my suggestion (replacing from l.42 to l.67)

```js
                let color;
                let icon;

                const detector = detectorsQualities.find((x) => x.id === id);

                // Run has the detector
                if (detector) {
                    const quality = detector.quality || null;

                    const actualQuality = detailsModel.isEditModeEnabled
                        ? detailsModel.runPatch.getDetectorQuality(id) || quality
                        : quality;

                    switch (actualQuality) {
                        case RunDetectorQualities.GOOD:
                            color = 'success';
                            icon = iconCheck();
                            break;
                        case RunDetectorQualities.BAD:
                            color = 'danger';
                            icon = iconX();
                            break;
                        // No quality specified for this detector
                        default:
                            color = 'gray-darker';
                            icon = iconBan();
                            break;
                    }
                // Run do not have the detector
                } else {
                    color = 'gray';
                    icon = iconBan();
                }
                
```

martinboulais (changes requested):

martinboulais (approved):

PR #1147, opened 2023-09-08
[O2B-989] filtering runs by updatedAt

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- NA

- add support for `/api/runs` endpoint to filter on `updatedAt`
- modify hook `resetDatabaseContent` - add function which modifies timestamps of four runs. It is added because runs from seeder has almost always the same timestamps so filtering by `updatedAt` is pointless.

Changes made to the database:
- NA

### DIFF
--- a/test/utilities/resetDatabaseContent.js
+++ b/test/utilities/resetDatabaseContent.js
@@ -12,9 +12,11 @@
  */
 
 const { database } = require('../../lib/application.js');
+const { setRunsTimestamps } = require('../testutils/setRunsTimestamps.js');
 
 exports.resetDatabaseContent = async () => {
     await database.dropAllTables();
     await database.migrate();
     await database.seed();
+    await setRunsTimestamps();
 };
--- a/test/utilities/resetDatabaseContent.js
+++ b/test/utilities/resetDatabaseContent.js
@@ -12,7 +12,7 @@
  */
 
 const { database } = require('../../lib/application.js');
-const { setRunsTimestamps } = require('../testutils/setRunsTimestamps.js');
+const { setRunsTimestamps } = require('./setRunsTimestamps.js');
 
 exports.resetDatabaseContent = async () => {
     await database.dropAllTables();
--- a/test/utilities/setRunsTimestamps.js
+++ b/test/utilities/setRunsTimestamps.js
@@ -1,0 +1,45 @@
+/**
+ * @license
+ * Copyright CERN and copyright holders of ALICE O2. This software is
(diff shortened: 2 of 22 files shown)


test/utilities/setRunsTimestamps.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+const setRunsTimestamps = async () => {
+    const dbPromises = runRunNumbersWithCustomizedTimestamps.map((runNumber) =>
+        sequelize.query(`UPDATE runs 
+            SET 
+                created_at=${getEscapedPastDate(runNumber)},
+                updated_at=${getEscapedPastDate(runNumber)} 
+            WHERE run_number = ${runNumber}`));
+
+    return Promise.all(dbPromises);
```
xsalonx, line 46: this way, because,
- update, upsert methods doesn't work
- creating new run breaks other tests

test/utilities/resetDatabaseContent.js
```diff
@@ -12,9 +12,11 @@
...
 
 const { database } = require('../../lib/application.js');
+const { setRunsTimestamps } = require('./setRunsTimestamps.js');
 
 exports.resetDatabaseContent = async () => {
     await database.dropAllTables();
     await database.migrate();
     await database.seed();
+    await setRunsTimestamps();
 };
```
martinboulais, line 22: A function named resetDatabaseContent should only reset database content to its default state, not do anything else.

test/utilities/setRunsTimestamps.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+const runRunNumbersWithCustomizedTimestamps = [1, 2, 3, 4];
+
+/**
+ * Set timestamps (created_at and updated_at) for runs with run_number specified in variable @see runRunNumbersWithCustomizedTimestamps
+ * according to formula `UPSERTED_TIMESTAMP = THIS_METHOD_CALL_TIME - run_number * 1000 * 60 * 30` (run_number * 30min)
+ * so there are time differences between the runs
+ * @return {Promise<Array<Object>>} information from db whether update was successful or not
+ */
+const setRunsTimestamps = async () => {
```
martinboulais, line 38: When you name function, think that this function name will be suggested in auto completion. 
`setRunsTimestamp` is way too generic to be exported as is, especially when it's a test function.

martinboulais (changes requested):

PR #1159, opened 2023-09-14
Attempt at fixing abseil CMake mess

ktf:


awegrzyn:
@ktf Could you explain why this is needed? As far as I remember `absl` is only build time dependency of `grpc` and Bookkeeping does not use it directly.

ktf:
I think without this we end up having an implicit dependency on absl when building QC, which might end up picking up different versions. That said, I think I found a different way of hiding absl in O2, so hopefully this is not needed.

awegrzyn:
Okay, Bookkeeping cmake config only requires `Protobuf` and `gRPC` so it should require `absl` when building QC. If it's the case, let us know and we'll fix the CMake.

PR #1193, opened 2023-10-16
[O2B-1025] - lhc periods table extraction

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- NA

- update `runs` related usecases

Changes made to the database:
- add table `lhc_periods`
- add model `LHCPeriod` referenced only by `Run` model
- replace column `runs`.`lhc_period` with `lhc_period_id` - reference to table `lhc_periods`
- migrate `lhc_period` data from `runs`
- update seeders


lib/database/adapters/LhcPeriodAdapter.js
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Converts the given entity object to a database object.
+     *
+     * @param {LhcPeriod} entityObject Object to convert.
+     * @returns {SequelizeLhcPeriod} Converted database object.
+     */
+    toDatabase(entityObject) {
+        return entityObject;
+    }
```
martinboulais, line 44: please explicit the fields here

lib/database/adapters/RunAdapter.js
```diff
@@ -239,7 +245,7 @@ class RunAdapter {
             pdpConfigOption: entityObject.pdpConfigOption,
             pdpTopologyDescriptionLibraryFile: entityObject.pdpTopologyDescriptionLibraryFile,
             tfbDdMode: entityObject.tfbDdMode,
-            lhcPeriod: entityObject.lhcPeriod,
+            lhcPeriod: { name: entityObject.lhcPeriod },
```
martinboulais, line 248: If at this point entityObject.lhcPeriod is a string, it means it's not in a valid state.
In the entity (service layer, which is, or should be, the core layer), the `lhcPeriod` is already an object => call the appropriate adapter (check for null first)

lib/database/migrations/20231011103554-create-lhc-period.js
```diff
@@ -0,0 +1,135 @@
...
+        }, { transaction });
+
+        const undoUpdateStmt = await prepareRunsUndoUpdateStatement({ queryInterface, transaction });
+        if (undoUpdateStmt) {
+            await sequelize.query(undoUpdateStmt, { transaction });
+        }
+
+        await queryInterface.removeColumn('runs', 'lhc_period_id', { transaction });
+        await queryInterface.dropTable('lhc_periods', { transaction });
+    }),
```
martinboulais, line 134: Why do you do something so complex? You should try to keep things simple.
1- Create the table
2- `raw` Use INSERT SELECT to fill the table using run's periods value (https://mariadb.com/kb/en/insert-select/)
3- Create the column in runs table
4- `raw` UPDATE ... JOIN to set the `lhc_period_id` in runs table (https://stackoverflow.com/questions/47546152/how-to-update-a-row-based-a-joined-table-in-mariadb)
5- Delete the `lhc_period` column

No need for complex intermediate data passed in js (takes a lot of memory for not much)

lib/database/models/index.js
```diff
@@ -60,6 +61,7 @@ module.exports = (sequelize) => {
         Subsystem: Subsystem(sequelize),
         Tag: Tag(sequelize),
         User: User(sequelize),
+        LHCPeriod: LHCPeriod(sequelize),
```
martinboulais, line 64: We don't use full capital, `LhcPeriod` (dumb or not we can discuss, but for now it's the convention)

lib/usecases/run/GetAllRunsUseCase.js
```diff
@@ -207,8 +207,24 @@ class GetAllRunsUseCase {
...
+                            required: true,
+                            attributes: [],
+                            where: {
+                                name: { [Op.in]: lhcPeriodsList },
+                            },
+                        },
+                    ],
+                });
+
+                filteringQueryBuilder.where('id').oneOf(...runsWithExpectedLHCPeriod.map(({ id }) => id));
```
martinboulais, line 225: You can probably apply directly the include on the filterQueryBuilder, as for the "or" case of tags.

martinboulais (changes requested):

lib/database/migrations/20231011103554-create-lhc-period.js
xsalonx: simplified

lib/database/models/index.js
xsalonx: renamed

lib/database/adapters/LhcPeriodAdapter.js
xsalonx: done

lib/usecases/run/GetAllRunsUseCase.js
xsalonx: right, modified

lib/database/adapters/RunAdapter.js
martinboulais: No. Here, `entityObject.lhcPeriod` must be an instance of `LhcPeriod` or null.

lib/database/migrations/20231011103554-create-lhc-period.js
```diff
@@ -106,28 +36,43 @@ module.exports = {
...
-        if (undoUpdateStmt) {
-            await sequelize.query(undoUpdateStmt, { transaction });
-        }
+        const updateRunsStmt = `
+            INSERT INTO runs(run_number, lhc_period)
+                SELECT r.run_number, p.name
+                FROM runs as r
+                LEFT JOIN lhc_periods as p
+                    ON p.id = r.lhc_period_id
+            ON DUPLICATE KEY UPDATE lhc_period = values(lhc_period)
```
martinboulais, line 73: Same comments as before

lib/database/migrations/20231011103554-create-lhc-period.js
```diff
@@ -106,28 +36,43 @@ module.exports = {
...
-        );
-        if (updateStmt) {
-            await queryInterface.sequelize.query(updateStmt, { transaction });
-        }
+        const insertPeriodsStmt = `
+            INSERT INTO lhc_periods(name)
+                SELECT DISTINCT lhc_period 
+                FROM runs
+                    WHERE lhc_period IS NOT NULL
+            ON DUPLICATE KEY UPDATE name = values(name)
```
martinboulais, line 44: You have to create the table first. I don't think your `ON DUPPLICATE KEY UPDATE` is a good idea. First the table has been created right before so you must not have dupplicate, and I think that you should almost NEVER update an ID

lib/database/migrations/20231011103554-create-lhc-period.js
```diff
@@ -106,28 +36,43 @@ module.exports = {
...
+        `;
+        await queryInterface.sequelize.query(insertPeriodsStmt);
+
+        const updateRunsDataStmt = `
+            INSERT INTO runs(run_number, lhc_period_id)
+                SELECT r.run_number, p.id
+                FROM runs as r
+                INNER JOIN lhc_periods as p
+                    ON p.name = r.lhc_period
+            ON DUPLICATE KEY UPDATE lhc_period_id = values(lhc_period_id)
```
martinboulais, line 54: Same comment, plus: why do you have an insert? This should be an update

lib/database/migrations/20231011103554-create-lhc-period.js
```diff
@@ -106,28 +36,43 @@ module.exports = {
...
         }, { transaction });
 
-        const updateStmt = await prepareRunsUpdateStatement(
-            await createAndGetPeriods({ queryInterface, transaction }),
-            { queryInterface, transaction },
-        );
-        if (updateStmt) {
-            await queryInterface.sequelize.query(updateStmt, { transaction });
-        }
+        const insertPeriodsStmt = `
```
martinboulais, line 39: Avoid arbitrary shortened names. We usually use `query` (SQL Query), I think statement is too generic

lib/database/migrations/20231011103554-create-lhc-period.js
martinboulais: See new comments

lib/database/repositories/LhcPeriodRepository.js
```diff
@@ -0,0 +1,34 @@
...
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+const {
+    models: {
+        LhcPeriod,
+    },
+} = require('../');
+const Repository = require('./Repository');
```
martinboulais, line 20: I think this fit on one line

lib/server/controllers/runs.controller.js
```diff
@@ -36,6 +36,19 @@ const {
 const { dtoValidator } = require('../utilities');
 const { ApiConfig } = require('../../config/index.js');
 
+/**
+ * Adjust Run entity to http api
```
martinboulais, line 40: I think `adapt` would be better than `adjust`

lib/server/services/run/RunService.js
```diff
@@ -172,7 +173,8 @@ class RunService {
      */
     async update(identifier, payload) {
         const { runPatch = {}, relations = {}, metadata = {} } = payload;
-        const { tagsTexts = null, eorReasons = null, runTypeName = null, userIdentifier = {}, detectorsQualities = [] } = relations;
+        const { tagsTexts = null, eorReasons = null, runTypeName = null,
```
martinboulais, line 176: Please update docblock

lib/usecases/run/EndRunUseCase.js
```diff
@@ -26,10 +26,13 @@ class EndRunUseCase {
...
         const { body, params } = dto;
 
+        const { lhcPeriod } = body;
+        delete body.lhcPeriod;
+
         // Bad variable name, the parameter expect a run number
         const { runId: runNumber } = params;
         try {
-            return { result: await runService.update({ runNumber }, { runPatch: body }) };
+            return { result: await runService.update({ runNumber }, { runPatch: body, relations: { lhcPeriod } }) };
```
martinboulais, line 35: Here, I think you should put lhcPeriodName
If you did not catch it, it means that a test should be added

test/lib/usecases/run/EndRunUseCase.test.js
```diff
@@ -62,7 +62,7 @@ module.exports = () => {
         expect(result.pdpConfigOption).to.equal('Repository hash');
         expect(result.pdpTopologyDescriptionLibraryFile).to.equal('production/production.desc');
         expect(result.tfbDdMode).to.equal('processing');
-        expect(result.lhcPeriod).to.equal('lhc22b');
+        chai.assert.deepStrictEqual(result.lhcPeriod, { id: 2, name: 'LHC22b' });
```
martinboulais, line 65: Please use `expect().to.eql()` instead

martinboulais (dismissed):

martinboulais (approved):

PR #1206, opened 2023-10-23
Public routes in development

xsalonx:
#### I DON'T have JIRA ticket
- [ ] tests are added

Notable changes for users:
- NA

- extract `isInTestMode` method from application to utilities and add similar `isInDevMode`
- use it instead of `process.env.NODE_ENV === ...` in a few places

Changes made to the database:
- NA


lib/server/routers/index.js
```diff
@@ -72,7 +73,7 @@ const inheritArgs = (parent, child) => {
 const bindRoute = (http, route, parentPath = '') => {
     const localPath = appendPath(parentPath, route.path, route.pathOption);
     if (route.method && route.controller) {
-        if (process.env.NODE_ENV === 'test') {
+        if (isInTestMode() || isInDevMode()) {
```
martinboulais, line 76: As I said, I disagree with this.
We need to be able to check that the routes are correctly blocked in development, and I even disagree to make them all public in test mode.
Please remove this change, but the refactoring is otherwise a good idea.

martinboulais (changes requested):

lib/server/routers/index.js
xsalonx: removed

martinboulais (approved):

PR #1217, opened 2023-10-30
[O2B-851] allow only admin to create and delete tags

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- only admin is authorised to create and delete tags

- NA

Changes made to the database:
- NA

### DIFF
--- a/lib/server/routers/tags.router.js
+++ b/lib/server/routers/tags.router.js
@@ -33,7 +33,6 @@ module.exports = {
         {
             method: 'post',
             controller: [rbacMiddleware([BkpRoles.ADMIN]), TagsController.createTag],
-            args: { public: true },
         },
         {
             path: ':tagId',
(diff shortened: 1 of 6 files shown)


graduta (changes requested):
Please update the tests accordingly and add new tests in the API folder

lib/server/routers/tags.router.js
```diff
@@ -32,7 +32,8 @@ module.exports = {
         },
         {
             method: 'post',
-            controller: TagsController.createTag,
+            controller: [rbacMiddleware([BkpRoles.ADMIN]), TagsController.createTag],
+            args: { public: true },
```
graduta, line 36: Why did you add `public true`?

graduta (changes requested):

lib/server/routers/tags.router.js
xsalonx: by accident

PR #1235, opened 2023-11-15
[O2B-1039] center of mass energy on periods page

xsalonx:
#### I have a JIRA ticket
- [ ] branch and/or PR name(s) include(s) JIRA ID
- [ ] PR labels are selected

Notable changes for users:
- TODO

- TODO

Changes made to the database:
- TODO


martinboulais (changes requested):

lib/database/models/typedefs/SequelizeLhcPeriodStatistics.js
```diff
@@ -15,6 +15,7 @@
  * @typedef SequelizeLhcPeriodStatistics
  *
  * @property {number} id
- * @property {number} avgEnergy
+ * @property {number|null} avgCenterOfMassEnergy
+ * @property {number[]|null} distinctEnergies
```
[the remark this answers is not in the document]
xsalonx: It's because some of periods do not have any PHYSICAL runs with defined lhc_beam_energy or pdp_beam_type so it inevitably returns null

lib/public/views/lhcPeriods/ActiveColumns/lhcPeriodsActiveColumns.js
```diff
@@ -20,9 +51,29 @@ export const lhcPeriodsActiveColumns = {
...
+                    ?.map((e) =>
+                        getClosestDefinedEnergy(
+                            e,
+                            distinctEnergiesProcessingConfiguration.values,
+                            distinctEnergiesProcessingConfiguration.acceptableEnergyMargin,
+                        ))
+                    .filter((value, index, array) => array.indexOf(value) === index)
+                    .map((energy) => Number(energy).toFixed(2))
+                    .sort()
+                    .join(', '))
```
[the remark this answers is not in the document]
xsalonx: ok, but there is not nested ternary (second question mark is null/undefined checker)

lib/database/migrations/20231025101611-create-lhc-periods-view.js
```diff
@@ -1,14 +1,28 @@
...
+    [`${CENTER_OF_MASS_ENERGY_FACTOR_KEY_PREFIX}_cosmic`]: 1,
+};
+
 const CREATE_PERIODS_VIEW = `
 CREATE OR REPLACE VIEW lhc_periods_statistics AS
 SELECT  
     p.id,
-    AVG(r.lhc_beam_energy) AS avgEnergy
+    AVG(r.lhc_beam_energy * center_of_mass_energy_factors.value) AS avg_center_of_mass_energy,
+    GROUP_CONCAT(DISTINCT(r.lhc_beam_energy)) AS distinct_energies
```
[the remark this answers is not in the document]
xsalonx: If not like this I see two other scenerios 
1. Put such a include in `LhcPeriodStatisticsService`
```
queryBuilder.include({
            association: 'lhcPeriod',
            include: [
                {
                    association: 'runs',
                    attributes: [
                        [
                            Sequelize.fn('GROUP_CONCAT', Sequelize.fn('DISTINCT', Sequelize.col('`lhcPeriod->runs`.lhc_beam_energy'))),
                            'distinctEnergies',
                        ],
                    ],
                },
            ],
        });

        queryBuilder.options.group = ['`LhcPeriodStatistics`.`id`'];
```
then adapt `distinctEnergies` (comma separated string) from `lhcPertiod` from `LhcPeriodStatisticsModel` 

2.  Put similar include (without distinct):
```
queryBuilder.include({
            association: 'lhcPeriod',
            include: [
                {
                    association: 'runs',
                    attributes: [
                        'lhcBeamEnergy',
                    ],
                },
            ],
        });
```
and then do unique on lhcBeamEnergies
___
Both are significantly more complex (in terms of computational complexity and code complexity) than current
Or maybe something else?

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
```diff
@@ -0,0 +1,93 @@
...
+ *
+ *      If consider all energies `E DEFINED IN wellDefinedEnergyLevels`
+ *      and take minimal absolute value of difference of lhcBeamEnergy and E
+ *      and `|lhcBeamEnergy - closestEnergy| <= acceptableMargin`
+ *      then the function returns `closestEnergy`
+ *      if there is no `closestEnergy` fulfilling all requirements then the function returns `lhcBeamEnergy`.
+ *
+ * @param {number} lhcBeamEnergy list of energies to be transformed
+ * @returns {string} closest energy in displayable format
+ */
```
martinboulais, line 68: I would rather name this function: `displayLhcBeamEnergy`. Docblock would be:

```js
/**
 * Display a given LHC beam energy
 * 
 * If the energy is close enough (within a given margin) to a well defined energy level, displays the energy level instead
 * 
 * @param {number} lhcBeamEnergy the energy to display
 * @return {string} the energy display
 */
```

Then if needed, you can create a function `displayLhcBeamEnergies` to display the list of energies. Here you have `lhcBeamEnergy list of energies`
```

File: lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js [hunk 5]
```diff
@@ -0,0 +1,93 @@
...
+
+const distinctEnergiesProcessingConfiguration = { // FOR NOW HARDCODED
+    wellDefinedEnergyLevelToDisplay: {
+        450: 450,
+        6800: 6800,
+        7000: 7000,
+        2680: '5360/2',
+    },
+    acceptableEnergyMargin: 0.01,
+};
```
martinboulais, line 24: If you don't need it yet, don't put it! Use constants when you can.
I would rather do:

```js
const WELL_DEFINED_ENERGY_LEVELS = [450, 2680, 6800, 7000]; // Sorted!
const ENERGY_LEVELS_CUSTOM_NAME = {2680: '5360/2'};
const ENERGY_LEVEL_MARGIN = 0.01;
const DISPLAY_PRECISION = 2;
```

Then: 
```js
getClosestEnergyLevel(lhcBeamEnergy) { // No need for the second argument, if it's needed at some point simply add it then
    for (const energyLevel of WELL_DEFINED_ENERGY_LEVELS) {
        if (lhcBeamEnergy < energyLevel - ENERGY_LEVEL_MARGIN) {
            return null; // next energy level is out of reach
        } else if (lhcBeamEnergy <= energyLevel + ENERGY_LEVEL_MARGIN) {
            return energyLevel;
        }
        // There may be other energy levels in reach, continue looping
    }
    return null;
}
```

Then to display, you simply do 
```js
const energyLevel = getClosestEnergyLevel(lhcBeamEnergy);
const energyToDisplay = energyLevel !== null ? energyLevel : lhcBeamEnergy;
energyToDisplay in ENERGY_LEVELS_CUSTOM_NAME ? ENERGY_LEVELS_CUSTOM_NAME[energyToDisplay] : energyToDisplay.toFixed(DISPLAY_PRECISION);
```

lib/database/migrations/20231025101611-create-lhc-periods-view.js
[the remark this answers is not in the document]
martinboulais: In terms of SQL efficiency I agree than doing the concat here is probably more optimized. But then, you will have to do a lot of split, which is probably worse.
In terms of code complexity, I think that doing group concat then split is worse than puting an include on a column, and you will have directly the right value.
Plus, what happens if at some point you want another property of run?

lib/database/migrations/20231025101611-create-lhc-periods-view.js
```diff
@@ -1,14 +1,26 @@
 'use strict';
 
+const BASIC_PARTICLES_PROPERTIES = { // 2 * sqrt(ATOMIC_MASS / ATOMIC_NUMBER)
```
martinboulais, line 3: Why basic?
Simply `PHYSICS_CONSTANTS` is enough for me. Or `PHYSICAL_CONSTANTS` as you named your table (not sure which one is correct, please check)

lib/database/models/typedefs/SequelizeLhcPeriodStatistics.js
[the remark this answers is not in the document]
martinboulais: If there is no physical runs, distingEnergies should be an empty array, not a null value.

martinboulais (changes requested):

lib/database/migrations/20231025101611-create-lhc-periods-view.js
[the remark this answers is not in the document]
xsalonx: refactored, but in different way that I proposed previously, because those solution were not feasible, due to sequelize problem (bug) with subQuery, limit and ordering..., added required attribute as literal

lib/database/models/typedefs/SequelizeLhcPeriodStatistics.js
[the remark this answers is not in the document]
xsalonx: When talking about type `LhcPeriodStatistics` then ok, but `SequelizeLhcPeriodStatistics`.`distinctEnergies` still might be null https://github.com/AliceO2Group/Bookkeeping/pull/1235#discussion_r1399165159

lib/public/app.css
```diff
@@ -653,3 +653,10 @@ label {
         grid-template-columns: repeat(3, 1fr);
     }
 }
+
+.center-of-mass-energy {
+    background-image: url('/assets/center-of-mass-energy.svg');
+    width: 40px;
+    height: 15.66px;
+    display: inline-block;
+}
```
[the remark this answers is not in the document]
xsalonx: refactored

lib/database/migrations/20231025101611-create-lhc-periods-view.js
xsalonx: refactored

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
xsalonx: refactored

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
xsalonx: applied

lib/database/utilities/QueryBuilder.js
```diff
@@ -470,6 +470,18 @@ class QueryBuilder {
...
     }
 
+    /**
+     * Add new attribute to result
+     * @param {string} attributeDefinition sql definition of attribute, e.g. function or subquery
+     * @param {string} attributeAlias alias for the attribute
+     * @return {QueryBuilder} this
+     */
+    includeAttribute(attributeDefinition, attributeAlias) {
+        this.options.attributes = this.options.attributes ?? { include: [] };
```
martinboulais, line 480: I think an `if` would more efficient and clear here

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
```diff
@@ -13,81 +14,50 @@
...
-export const getClosestDefinedEnergyLevel =
-    (lhcBeamEnergy, wellDefinedEnergyLevels) => {
-        const definedEnergyValues = Object.keys(wellDefinedEnergyLevels);
-        const closestEnergyLevel = definedEnergyValues.reduce(
-            (acc, current) => Math.abs(current - lhcBeamEnergy) < Math.abs(acc - lhcBeamEnergy) ? current : acc,
-            definedEnergyValues[0],
-        );
-        return closestEnergyLevel;
-    };
+const WELL_DEFINED_ENERGY_LEVELS = [450, 2680, 6800, 7000].sort((a, b) => a - b); // Sorted!
```
martinboulais, line 17: You defined them as sorted, no need to sort it again
You can put the comment saying that the array is expected to be sorted though

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
```diff
@@ -13,81 +14,50 @@
...
- * @param {number} acceptableEnergyMargin minimal energy difference to keep lhcBeamEnergy as is
- * @param {number} notChangedEnergyPrecision number of significant decimal places
- * @returns {string} lhcBeamEnergy or closest energy in displayable format
- */
-const _distinctLhcBeamEnergyToDisplay =
-    (lhcBeamEnergy, closestEnergyLevel, wellDefinedEnergyLevels, acceptableEnergyMargin, notChangedEnergyPrecision = 2) =>
-        Math.abs(closestEnergyLevel - lhcBeamEnergy) <= acceptableEnergyMargin
-            ? wellDefinedEnergyLevels(closestEnergyLevel)
-            : lhcBeamEnergy.toFixed(notChangedEnergyPrecision).toString();
+function getClosestEnergyLevel(lhcBeamEnergy) {
```
martinboulais, line 22: Missing docblock

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
```diff
@@ -13,81 +14,50 @@
...
+ * Format lhc beam distinct energies, optionaly replacing some of them
+ * by well defined energy levels @see displayLhcBeamEnergy
+ * @param {number[]} distinctEnergies list of energies to be displayed
+ * @returns {Component} formatted list of distinct energies
  */
 export const formatDistinctLhcBeamEnergies = (distinctEnergies) =>
     distinctEnergies?.length > 0
         ? h('', distinctEnergies
-            ?.map(distinctLhcBeamEnergyToDisplay)
+            ?.map(displayLhcBeamEnergy)
```
martinboulais, line 59: Optional chaining is not needed here, and it's too recent to be used (we can discuss this afternoon the minimum version we target)

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
```diff
@@ -13,81 +14,50 @@
...
             .filter((value, index, array) => array.indexOf(value) === index) // Unique
```
martinboulais, line 60: I don't know which one is the more optimized, but I would rather create a `Set` and go back to the array

martinboulais, line 60: I don't know which one is the more optimized, but I would rather create a `Set` and go back to the array

lib/database/adapters/LhcPeriodStatisticsAdapter.js
```diff
@@ -30,8 +30,9 @@ class LhcPeriodStatisticsAdapter {
      * @returns {LhcPeriodStatistics} Converted entity object.
      */
     toEntity(databaseObject) {
-        const { id, avgCenterOfMassEnergy, distinctEnergies, lhcPeriod } = databaseObject;
-        const entity = { id, avgCenterOfMassEnergy, distinctEnergies };
+        const { id, avgCenterOfMassEnergy, lhcPeriod } = databaseObject;
+        const distinctEnergies = databaseObject.get('distinctEnergies');
+        const entity = { id, avgCenterOfMassEnergy, distinctEnergies: distinctEnergies?.split(',').map((energy) => Number(energy)) ?? [] };
```
martinboulais, line 35: Why don't you want to simply include runs's column you are interested in?

lib/database/repositories/LhcPeriodStatisticsRepository.js
```diff
@@ -25,6 +36,29 @@ class LhcPeriodStatisticsRepository extends Repository {
...
     }
+
+    /**
+     * Returns a specific entity.
+     *
+     * @param {QueryBuilder} [queryBuilder] the find query (see sequelize findAll options) or a find query builder
+     * @returns {Promise<Object>|null} Promise object representing the full mock data
+     */
+    async findOne(queryBuilder = dataSource.createQueryBuilder()) {
+        queryBuilder.includeAttribute(...DISTINCT_ENERGIES_IN_STATISTICS_ATTRIBUTE);
```
martinboulais, line 47: This do not belong here: you expect find one to fetch what you pass as query builder, not to take initiatives

lib/database/repositories/LhcPeriodStatisticsRepository.js
```diff
@@ -25,6 +36,29 @@ class LhcPeriodStatisticsRepository extends Repository {
...
+
+    /**
+     * Returns all entities.
+     *
+     * @param {QueryBuilder} [queryBuilder] The QueryBuilder to use.
+     * @returns {Promise} Promise object representing the full mock data
+     */
+    async findAndCountAll(queryBuilder = dataSource.createQueryBuilder()) {
+        queryBuilder.set('distinct', true);
+        queryBuilder.includeAttribute(...DISTINCT_ENERGIES_IN_STATISTICS_ATTRIBUTE);
```
martinboulais, line 59: Same, both for set and includeAttribute

martinboulais (changes requested):

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
xsalonx: /**
 * Reason for calculation of runs lhc beam distinct energies this way (using literals) is:
 * There is transitive association: `runs -> lhc_periods -> lhc_periods_statistics`.
 * Including runs in queries below causes that sequelize build internaly subquery to fetch data.
 * This subquery is an `SELECT ... FROM lhc_periods_statistics`,
 * but after adding ordering by `lhcPeriod.name` sequelize puts the `ORDER BY` clause inside that subquery,
 * what leads to error, because lhcPeriod is not in scope of the subquery.
 * Mentioned can be avoid by providing flag:  `subQuery: false`,
 * but this breaks limit clause, so pagination.
 * There is related issue for reference: https://github.com/sequelize/sequelize/issues/11983
 */

lib/database/adapters/LhcPeriodStatisticsAdapter.js
xsalonx: /**
 * Reason for calculation of runs lhc beam distinct energies this way (using literals) is:
 * There is transitive association: `runs -> lhc_periods -> lhc_periods_statistics`.
 * Including runs in queries below causes that sequelize build internaly subquery to fetch data.
 * This subquery is an `SELECT ... FROM lhc_periods_statistics`,
 * but after adding ordering by `lhcPeriod.name` sequelize puts the `ORDER BY` clause inside that subquery,
 * what leads to error, because lhcPeriod is not in scope of the subquery.
 * Mentioned can be avoid by providing flag:  `subQuery: false`,
 * but this breaks limit clause, so pagination.
 * There is related issue for reference: https://github.com/sequelize/sequelize/issues/11983
 */

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
xsalonx: yes, sets are faster: https://medium.com/@jakubsynowiec/unique-array-values-in-javascript-7c932682766c

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
xsalonx: optional chaining removed

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
xsalonx: added

lib/public/views/lhcPeriods/format/formatDistinctLhcBeamEnergies.js
xsalonx: same

martinboulais: This is what I thought

martinboulais (approved):

PR #1236, opened 2023-11-16
[O2B-1043] Add time filter

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- User can now chose the date range on which statistics must be applied

- A time range picker has been created

### DIFF
--- a/lib/public/components/Filters/common/FilterModel.js
+++ b/lib/public/components/Filters/common/FilterModel.js
@@ -14,6 +14,7 @@ import { Observable } from '/js/src/index.js';
 
 /**
  * Model storing the state of a given filter
+ * @abstract
  */
 export class FilterModel extends Observable {
     /**
@@ -27,27 +28,32 @@ export class FilterModel extends Observable {
 
     /**
      * Reset the filter to its initial state
+     *
      * @return {void}
+     * @abstract
      */
     reset() {
+        throw new Error('Abstract function call');
     }
 
     /**
      * States if the filter has been filled with a valid value
      *
      * @return {boolean} true if the filter is filled
+     * @abstract
      */
     get isEmpty() {
-        return true;
+        throw new Error('Abstract function call');
     }
 
     /**
      * Returns the normalized value of the filter, that can be used as URL parameter
      *
      * @return {string|number|object|array|null} the normalized value
+     * @abstract
      */
     get normalized() {
-        return null;
+        throw new Error('Abstract function call');
     }
 
     /**
--- a/lib/public/components/Filters/common/FilterModel.js
+++ b/lib/public/components/Filters/common/FilterModel.js
@@ -1,0 +1,61 @@
+/**
+ * @license
+ * Copyright CERN and copyright holders of ALICE O2. This software is
+ * distributed under the terms of the GNU General Public License v3 (GPL
+ * Version 3), copied verbatim in the file "COPYING".
+ *
+ * See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+import { Observable } from '/js/src/index.js';
+
+/**
+ * Model storing the state of a given filter
+ */
+export class FilterModel extends Observable {
+    /**
+     * Constructor
+     */
+    constructor() {
+        super();
+
+        this._visualChange$ = new Observable();
+    }
+
+    /**
+     * Reset the filter to its initial state
+     * @return {void}
+     */
+    reset() {
+    }
+
+    /**
+     * States if the filter has been filled with a valid value
+     *
+     * @return {boolean} true if the filter is filled
+     */
+    get isEmpty() {
+        return true;
+    }
+
+    /**
+     * Returns the normalized value of the filter, that can be used as URL parameter
+     *
+     * @return {string|number|object|array|null} the normalized value
+     */
+    get normalized() {
+        return null;
+    }
+
+    /**
+     * Returns the observable notified any time there is a visual change which has no impact on the actual filter value
+     *
+     * @return {Observable} the observable
+     */
+    get visualChange$() {
+        return this._visualChange$;
+    }
+}
--- a/lib/public/components/Filters/common/filters/timeRangeFilter.js
+++ b/lib/public/components/Filters/common/filters/timeRangeFilter.js
@@ -19,17 +19,13 @@ import { getLocaleDateAndTime } from '../../../../utilities/dateUtils.js';
  * Format the current timestamp range display
  *
  * @param {Partial<Period>} the current period
- * @return {vnode|string} the resulting component
+ * @return {Component} the resulting component
  */
 const formatTimeRange = ({ from, to }) => {
     let parts = [];
 
     // eslint-disable-next-line require-jsdoc
     const formatTimestamp = (timestamp) => {
-        if (timestamp === undefined) {
-            return '-';
-        }
-
         const { date, time } = getLocaleDateAndTime(timestamp);
         return h('.flex-column.g1', [h('.f6', date), h('', time)]);
     };
--- a/lib/public/components/Filters/common/filters/timeRangeFilter.js
+++ b/lib/public/components/Filters/common/filters/timeRangeFilter.js
@@ -1,0 +1,83 @@
+/**
+ *  @license
+ *  Copyright CERN and copyright holders of ALICE O2. This software is
+ *  distributed under the terms of the GNU General Public License v3 (GPL
+ *  Version 3), copied verbatim in the file "COPYING".
+ *
+ *  See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ *  In applying this license CERN does not waive the privileges and immunities
+ *  granted to it by virtue of its status as an Intergovernmental Organization
+ *  or submit itself to any jurisdiction.
+ */
+import { dropdown } from '../../../common/popover/dropdown.js';
+import { h } from '/js/src/index.js';
+import { dateTimeInput } from '../../../common/form/inputs/dateTimeInput.js';
+import { getLocaleDateAndTime } from '../../../../utilities/dateUtils.js';
+
+/**
+ * Format the current timestamp range display
+ *
+ * @param {Partial<Period>} the current period
+ * @return {vnode|string} the resulting component
+ */
+const formatTimeRange = ({ from, to }) => {
+    let parts = [];
+
+    if (from === null && to === null) {
+        return '-';
+    }
+
+    // eslint-disable-next-line require-jsdoc
+    const formatTimestamp = (timestamp) => {
+        if (timestamp === undefined) {
+            return '-';
+        }
+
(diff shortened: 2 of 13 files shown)


lib/public/components/Filters/common/filters/timeRangeFilter.js
```diff
@@ -0,0 +1,83 @@
...
+
+    if (from === undefined && to === undefined) {
+        parts = '-';
+    } else if (from === undefined) {
+        parts = [label('Before'), formatTimestamp(to)];
+    } else if (to === undefined) {
+        parts = [label('After'), formatTimestamp(from)];
+    } else {
+        parts = [label('From'), formatTimestamp(from), label('to'), formatTimestamp(to)];
+    }
```
xsalonx, line 52: I think it can be achieved much shorter (if you don't care about zeroes): 
```
const parts = [
        ... from && [label('After'), formatTimestamp(from) || []],
        ... to && [label('Before'), formatTimestamp(to) || []],
];
```

xsalonx: another think is taking care about null values

lib/public/components/common/form/inputs/DateTimeInputModel.js
```diff
@@ -0,0 +1,316 @@
...
+        this._max = max;
+    }
+
+    /**
+     * Returns the given date formatted in two parts, YYYY-MM-DD and HH:MM to fill in HTML date and time inputs
+     *
+     * @param {number} timestamp the timestamp (ms) to format
+     *
+     * @return {{date: string, time: string}} the date expression to use as HTML input values
+     */
```
xsalonx, line 80: I think the exact format should be specified within typdef (https://github.com/AliceO2Group/Bookkeeping/pull/1236/files#diff-f12fbcc953abe82f5d9bce9e799eac8d90fc7bad60c256e6ce0bb1ea2b2ee962R19-R22) 
and if {date: string, time: string} == DateTimeInputRawData
```
@return {DateTimeInputRawData}
```

lib/public/components/common/form/inputs/DateTimeInputModel.js
```diff
@@ -0,0 +1,316 @@
...
+
+    /**
+     * Update the inputs raw values
+     *
+     * @param {Partial<DateTimeInputRawData>} patch the input values patch
+     * @return {void}
+     */
+    update({ date, time }) {
+        if (date !== undefined) {
+            this._raw.date = date;
```
xsalonx, line 107: what if null

lib/public/components/common/form/inputs/DateTimeInputModel.js
```diff
@@ -0,0 +1,316 @@
...
+        this._visualChange$ = new Observable();
+
+        this._isRequired = required;
+        this._min = min;
+        this._max = max;
+    }
+
+    /**
+     * Returns the given date formatted in two parts, YYYY-MM-DD and HH:MM to fill in HTML date and time inputs
+     *
```
xsalonx, line 76: Some docs starts from verb in third person (Returns) some not (Return)

lib/public/components/common/form/inputs/DateTimeInputModel.js
```diff
@@ -0,0 +1,316 @@
...
+     */
+    set max(value) {
+        this._max = value;
+    }
+
+    /**
+     * Returns the min values to apply to the inputs
+     *
+     * @return {(Partial<{date: string, time: string}>|string)} the min values
+     */
```
xsalonx, line 266: I think it cannot return string and is it intentional not to use DateTimeInputRawData ?

lib/public/components/common/form/inputs/DateTimeInputModel.js
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        return ret;
+    }
+
+    /**
+     * Returns the max values to apply to the inputs
+     *
+     * @return {(Partial<{date: string, time: string}>|string)} the max values
+     */
```
xsalonx, line 294: same as previous

lib/public/components/Filters/common/filters/timeRangeFilter.js
```diff
@@ -0,0 +1,83 @@
...
+ * Format the current timestamp range display
+ *
+ * @param {Partial<Period>} the current period
+ * @return {vnode|string} the resulting component
+ */
+const formatTimeRange = ({ from, to }) => {
+    let parts = [];
+
+    if (from === null && to === null) {
+        return '-';
```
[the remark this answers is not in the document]
martinboulais: I cleaned the code. From and to can not be null, because it's a partial period: they are either a number or `undefined` which makes sense if the period is openened on one/both side.

lib/public/components/Filters/common/filters/timeRangeFilter.js
```diff
@@ -0,0 +1,83 @@
...
+import { dropdown } from '../../../common/popover/dropdown.js';
+import { h } from '/js/src/index.js';
+import { dateTimeInput } from '../../../common/form/inputs/dateTimeInput.js';
+import { getLocaleDateAndTime } from '../../../../utilities/dateUtils.js';
+
+/**
+ * Format the current timestamp range display
+ *
+ * @param {Partial<Period>} the current period
+ * @return {vnode|string} the resulting component
```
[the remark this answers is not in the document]
martinboulais: You're right, component

lib/public/components/Filters/common/filters/timeRangeFilter.js
```diff
@@ -0,0 +1,83 @@
...
+
+        const { date, time } = getLocaleDateAndTime(timestamp);
+        return h('.flex-column.g1', [h('.f6', date), h('', time)]);
+    };
+
+    // eslint-disable-next-line require-jsdoc
+    const label = (content) => h('.gray-darker', content);
+
+    if (from === undefined && to === undefined) {
+        parts = '-';
```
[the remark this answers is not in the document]
martinboulais: Replied here

lib/public/components/Filters/common/filters/timeRangeFilter.js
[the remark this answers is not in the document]
xsalonx: ok

lib/public/components/Filters/common/filters/timeRangeFilter.js
```diff
@@ -0,0 +1,83 @@
...
+
+    if (from === null && to === null) {
+        return '-';
+    }
+
+    // eslint-disable-next-line require-jsdoc
+    const formatTimestamp = (timestamp) => {
+        if (timestamp === undefined) {
+            return '-';
+        }
```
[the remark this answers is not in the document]
martinboulais: Yes you're right

lib/public/components/Filters/common/filters/timeRangeFilter.js
martinboulais: With your solution, we are missing the case where both are defined.
This can be handled with a ternary inside what you proposed, but I think that this would make the code unclear, don't you think?

Something like:
```js
const parts = [
        ... from !== undefined && [label(to !== undefined ? 'From' : 'After'), formatTimestamp(from)],
        ... to !== undefined && [label(from !== undefined ? 'to' : 'Before'), formatTimestamp(to)],
];
```

lib/public/components/common/form/inputs/DateTimeInputModel.js
martinboulais: Absolutely

lib/public/components/common/form/inputs/DateTimeInputModel.js
martinboulais: It can't be null (`Partial<DateTimeInputRawData>`)

lib/public/components/common/form/inputs/DateTimeInputModel.js
martinboulais: Returns is the right one (I suppose)

lib/public/components/common/form/inputs/DateTimeInputModel.js
martinboulais: Same reply

lib/public/components/Filters/common/filters/timeRangeFilter.js
xsalonx: ok, I missed that labels differ. If altering them is important for some reason (e.g. it's nicer) then let it be as is

lib/public/components/common/form/inputs/DateTimeInputModel.js
xsalonx: ok

lib/public/components/common/form/inputs/DateTimeInputModel.js
martinboulais: You're right about the string.
Yes it is intentional, because yes it happens to have the same properties but it's not the same semantic, it's not inputs raw data

lib/public/components/common/form/inputs/DateTimeInputModel.js
xsalonx: yeah, I mentioned it just to keep it consistent

lib/public/components/common/form/inputs/DateTimeInputModel.js
xsalonx: so as not the same maybe another typedef, because they looks quite similar at the first glance, at least for me it was a bit misleading

lib/public/components/Filters/common/FilterModel.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Returns the normalized value of the filter, that can be used as URL parameter
+     *
+     * @return {string|number|object|array|null} the normalized value
+     */
+    get normalized() {
+        return null;
+    }
```
xsalonx, line 51: I have doubts about this method name:
1. If it is intended as general purpose accessor (getter) of internally stored values (with maybe some not serious adaptation) it might be just getValue (or something similar)
2. If it is intended to return values already suitable to be injected into search query params, it might be `toSeachQueryFormat` (or sth else)

xsalonx (changes requested):

lib/public/components/common/form/inputs/DateTimeInputModel.js
martinboulais: TBH it's used only here, I think it would be overkill to create a typedef for this. If it need to be used anywhere else, I would extract a typedef.

About the typedef: they describe the structure of something specific, not just a structure. You should never reuse the structure just because it maches your type.

lib/public/components/Filters/common/FilterModel.js
martinboulais: I have put this because it's a common function in every filter that return the filter's data in a normalized way. It's not really the value (for example value can be a date, but normalized will be a timestamp). It's not only search query format, because this can be an object, or nested objects... Not sure if I'm clear

lib/public/components/common/form/inputs/DateTimeInputModel.js
xsalonx: ok

xsalonx (approved):

PR #1240, opened 2023-11-17
[O2B-1045] nav bar adjustment

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Replace nav bar `LHC Fills` tab with `LHC` dropdown with two options:
    - `LHC Fills`
    - `LHC Periods`

- use only one method in model to toggle nav bar dropdowns
- change links id in https://github.com/AliceO2Group/Bookkeeping/pull/1240/files#diff-90f47b96a0752a7ea8dbe41825baf3804e1228ab28f024b599524e0d0ea2c551R47-R49, because of test failure after adding `LHC Fills` as `pageLink`

Changes made to the database:
- NA

### DIFF
--- a/lib/public/Model.js
+++ b/lib/public/Model.js
@@ -252,29 +252,12 @@ export default class Model extends Observable {
     }
 
     /**
-     * Toggle navbar overview dropdown
-     * @returns {undefined}
+     * Toggle navbar dropdown
+     * @param {string} key referencing dropdown to be toggled
+     * @returns {void}
      */
-    toggleOverviewDropdown() {
-        this.dropdownMenu = this.dropdownMenu === 'overview' ? false : 'overview';
-        this.notify();
-    }
-
-    /**
-     * Toggle navbar account dropdown
-     * @returns {undefined}
-     */
-    toggleAccountDropdown() {
-        this.dropdownMenu = this.dropdownMenu === 'account' ? false : 'account';
-        this.notify();
-    }
-
-    /**
-     * Toggle navbar LHC dropdown
-     * @returns {undefined}
-     */
-    toggleLHCDropdown() {
-        this.dropdownMenu = this.dropdownMenu === 'LHC' ? false : 'LHC';
+    toggleDropdownMenu(key) {
+        this.dropdownMenu = this.dropdownMenu === key ? null : key;
         this.notify();
     }
 
--- a/lib/public/Model.js
+++ b/lib/public/Model.js
@@ -270,6 +270,15 @@ export default class Model extends Observable {
     }
 
     /**
+     * Toggle navbar LHC dropdown
+     * @returns {undefined}
+     */
+    toggleLHCDropdown() {
+        this.dropdownMenu = this.dropdownMenu === 'LHC' ? false : 'LHC';
+        this.notify();
+    }
+
+    /**
      * Clears all navbar dropdowns
      * @returns {undefined}
      */
--- a/lib/public/Model.js
+++ b/lib/public/Model.js
@@ -266,7 +266,7 @@ export default class Model extends Observable {
      * @returns {undefined}
      */
     clearDropdownMenu() {
-        this.dropdownMenu = false;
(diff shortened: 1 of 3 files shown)


lib/public/Model.js
```diff
@@ -269,6 +269,15 @@ export default class Model extends Observable {
         this.notify();
     }
 
+    /**
+     * Toggle navbar LHC dropdown
+     * @returns {undefined}
+     */
+    toggleLHCDropdown() {
```
martinboulais, line 276: I think here you can extract a common toggle function, to which you simply pass a key

lib/public/components/NavBar/index.js
```diff
@@ -81,7 +81,17 @@ const navBar = (model) => {
             pageTab(model, 'home', 'Home'),
             pageTab(model, 'log-overview', 'Log Entries'),
             pageTab(model, 'env-overview', 'Environments'),
-            pageTab(model, 'lhc-fill-overview', 'LHC Fills'),
+            h(`.navbar-dropdown${model.dropdownMenu === 'LHC' ? '.dropdown-open' : ''}`, [
```
martinboulais, line 84: Several things here:
First, I think you can extract a `subMenu` or `dropdownSubMenu` component.
Then, you can probably use `dropdown` utitlity to clean this out.
Then, `pageLink` can probably use `frontLink`/`absoluteFrontLink`

martinboulais (changes requested):

lib/public/Model.js
xsalonx: refactored

lib/public/components/NavBar/index.js
xsalonx: refactored

lib/public/Model.js
```diff
@@ -252,29 +252,13 @@ export default class Model extends Observable {
...
-     * Clears all navbar dropdowns
-     * @returns {undefined}
+     * Toggle navbar dropdown
+     * @param {string|null} key referencing dropdown to be toggled,
+     * default value is null, which causes closing all dropdowns
+     * @returns {void}
      */
-    clearDropdownMenu() {
-        this.dropdownMenu = false;
+    toggleDropdownMenu(key = null) {
```
martinboulais, line 260: No, toggle is different from clear. Please keep the clear function separated, and do not allow for null key.

lib/public/components/NavBar/index.js
```diff
@@ -66,13 +66,32 @@ const aliFlpLinkTab = (aliFlpIndexPageUrl) => absoluteFrontLink(
 
 const btnGroupsDelimiter = h('.mh1', { style: { 'border-left': '1px solid var(--color-gray-dark)' } });
 
+/**
+ * Build navigation bar dropdown menu
+ * @param {Model} model model
+ * @param {string} key by which the model references dropdown menu
+ * @param {string|icon} dropdownButtonContent content of button which openes dropdown menu
+ * @param {Component[]} entries entries of the menu
+ * @param {object} [options.menuStyle] style for dropdown
```
martinboulais, line 75: it's rather attributes than style no?

lib/public/components/NavBar/index.js
martinboulais: What I meant by `Then, pageLink can probably use frontLink/absoluteFrontLink` is that:
You should have `pageLink` and `absoluteLink` and avoid to always have the construction of `?page=`
I would rather pass an `onClick` callback than the full model.

lib/public/components/NavBar/index.js
xsalonx: right

lib/public/Model.js
xsalonx: clear function restored

lib/public/components/NavBar/index.js
xsalonx: refactored

martinboulais (approved):

PR #1241, opened 2023-11-17
[O2B-1048] add year column

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- add year column for LHC Periods page

- NA

Changes made to the database:
- modify lhc_periods_statistics view

### DIFF
--- a/lib/database/migrations/20231025101611-create-lhc-periods-view.js
+++ b/lib/database/migrations/20231025101611-create-lhc-periods-view.js
@@ -4,7 +4,8 @@ const CREATE_PERIODS_VIEW = `
 CREATE OR REPLACE VIEW lhc_periods_statistics AS
 SELECT  
     p.id,
-    AVG(r.lhc_beam_energy) AS avgEnergy
+    AVG(r.lhc_beam_energy) AS avgEnergy,
+    if(name like 'LHC%', substring(name, 4, 2) + 2000, null) as year
 FROM lhc_periods AS p
 LEFT JOIN  runs as r
     ON r.lhc_period_id = p.id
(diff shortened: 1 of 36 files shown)


lib/database/migrations/20231025101611-create-lhc-periods-view.js
```diff
@@ -4,7 +4,8 @@ const CREATE_PERIODS_VIEW = `
 CREATE OR REPLACE VIEW lhc_periods_statistics AS
 SELECT  
     p.id,
-    AVG(r.lhc_beam_energy) AS avgEnergy
+    AVG(r.lhc_beam_energy) AS avgEnergy,
+    if(name like 'LHC%', substring(name, 4, 2) + 2000, null) as year
```
martinboulais, line 8: NEVER update a migration! Always create a new one.
If this migration has already been played, it fill never be played again.
I'm really thinking if this kind of column is worth storing in the database? Why not either create a virtual column or do in directly in the frontend?

martinboulais (changes requested):

lib/database/migrations/20231025101611-create-lhc-periods-view.js
xsalonx: This is a view we agreed to build step by step and as it's not merged to main it can be changed...
Having year here instead of on fronted or as virtual field will make filtering, sorting (and maybe other stuff) much simpler and generic

martinboulais: For filtering you simply do the equal/like in the filtering, for sorting it changes nothing?
I really think here you want the facility of making the model being the same as the view (front-end) which is not a good practice. Keeping the model as simple as possible is quite often the best choice.
Here you are adding business logic in the database, which you should never do.
If needed, you can have a year column in the database, and in the service layer you extract the value when saving an entity.

xsalonx: refactored:
- year calculated on frontend
- filtering and sorting handled as special cases inside service

martinboulais (approved):

PR #1248, opened 2023-11-23
[O2B-1049] add beam type column

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Add `Beam Type` column to `Lhc Periods` page

- add `beamType` attribute to LhcPeriodStatistics API
- small refactor of `QueryBuilder#includeAttribute`

Changes made to the database:
- NA


lib/database/adapters/LhcPeriodStatisticsAdapter.js
```diff
@@ -32,7 +32,13 @@ class LhcPeriodStatisticsAdapter {
     toEntity(databaseObject) {
         const { id, avgCenterOfMassEnergy, lhcPeriod } = databaseObject;
         const distinctEnergies = databaseObject.get('distinctEnergies');
-        const entity = { id, avgCenterOfMassEnergy, distinctEnergies: distinctEnergies?.split(',').map((energy) => Number(energy)) ?? [] };
+        const beamType = databaseObject.get('beamType');
```
martinboulais, line 35: Why using `.get` ?

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
```diff
@@ -28,15 +28,28 @@ const { NotFoundError } = require('../../errors/NotFoundError');
...
+};
+
+const PDP_BEAM_TYPES_IN_STATISTICS_ATTRIBUTE = {
+    query: `(SELECT r.pdp_beam_type
+        FROM runs AS r
+        WHERE r.lhc_period_id = lhcPeriod.id 
+            AND r.definition = 'PHYSICS'
+            AND r.pdp_beam_type IS NOT NULL
+            AND r.pdp_beam_type != 'cosmic'
+        LIMIT 1
```
martinboulais, line 48: Why limit 1 ?

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
```diff
@@ -28,15 +28,28 @@ const { NotFoundError } = require('../../errors/NotFoundError');
...
+    alias: 'distinctEnergies',
+};
+
+const PDP_BEAM_TYPES_IN_STATISTICS_ATTRIBUTE = {
+    query: `(SELECT r.pdp_beam_type
+        FROM runs AS r
+        WHERE r.lhc_period_id = lhcPeriod.id 
+            AND r.definition = 'PHYSICS'
+            AND r.pdp_beam_type IS NOT NULL
+            AND r.pdp_beam_type != 'cosmic'
```
martinboulais, line 47: Why? (maybe a link to the definition of what physicists expects? Or a comment on why do you need this)

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
```diff
@@ -28,15 +28,28 @@ const { NotFoundError } = require('../../errors/NotFoundError');
  * but this breaks limit clause, so pagination.
  * There is related issue for reference: https://github.com/sequelize/sequelize/issues/11983
  */
-const DISTINCT_ENERGIES_IN_STATISTICS_ATTRIBUTE = [
-    `(SELECT GROUP_CONCAT(DISTINCT(r.lhc_beam_energy))
+const DISTINCT_ENERGIES_IN_STATISTICS_ATTRIBUTE = {
+    query: `(SELECT GROUP_CONCAT(DISTINCT(r.lhc_beam_energy))
         FROM runs AS r
         WHERE r.lhc_period_id = lhcPeriod.id 
             AND r.definition = 'PHYSICS'
```
martinboulais, line 35: You have a `GROUP_CONCAT` without group by ?

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
```diff
@@ -66,7 +79,8 @@ class LhcPeriodStatisticsService {
             throw new BadParameterError('Can not find without LHC Period id or name');
         }
 
-        queryBuilder.includeAttribute(...DISTINCT_ENERGIES_IN_STATISTICS_ATTRIBUTE);
+        queryBuilder.includeAttribute(DISTINCT_ENERGIES_IN_STATISTICS_ATTRIBUTE);
+        queryBuilder.includeAttribute(PDP_BEAM_TYPES_IN_STATISTICS_ATTRIBUTE);
```
martinboulais, line 83: I think here you should name the constant `LHC_PERIOD_STATISTICS_DISTINCT_ENERGIES_SUBQUERY` and put the alias here, constant only for the query itself

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
```diff
@@ -118,6 +132,8 @@ class LhcPeriodStatisticsService {
                     expression = (sequelize) => sequelize.col('lhcPeriod.name');
                 } else if (property === 'year') {
                     expression = (sequelize) => sequelize.literal('SUBSTRING(lhcPeriod.name, 4, 2)');
+                } else if (property === 'beamType') {
+                    expression = (sequelize) => sequelize.literal(PDP_BEAM_TYPES_IN_STATISTICS_ATTRIBUTE.query);
                 }
```
martinboulais, line 137: Maybe use a switch at this point, or even better a `switchCase`

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
```diff
@@ -146,9 +162,16 @@ class LhcPeriodStatisticsService {
                     { years },
                 );
             }
+            if (beamTypes) {
+                queryBuilder.literalWhere(
+                    `${PDP_BEAM_TYPES_IN_STATISTICS_ATTRIBUTE.query} IN (:beamTypes)`,
```
martinboulais, line 167: If this works, it is probably some highly un-optmized query! I suppose that you have to run the subquery for every rows in the table to then run the IN

martinboulais (changes requested):

lib/database/adapters/LhcPeriodStatisticsAdapter.js
xsalonx: the same reason, why use get for 'distinctEnergies'
it is not filed defined in model, so it is not accessible by simple object destruction

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
xsalonx: It's assumed that lhc_period is period of some common parameters of runs (PHYSICS), e.g. the same beam type. If you see data in db, indeed, all PHYSICS runs belonging to the one lhc periods have the same pdp_beam_type.
Because some might be null (errors, whatever), there is `NOT NULL` condition
Because need only one portion of information about beam type, there is `LIMIT 1`

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
xsalonx: continuing [https://github.com/AliceO2Group/Bookkeeping/pull/1248#discussion_r1405779940] there is one run which is PHYSICS and has `cosmic` beam type and it's not something that should be presented

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
xsalonx: yes, because `GROUP_CONCAT` is applied only in scope of subquery, where all records are merged, so there is no need of `CROUP BY` clause

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
xsalonx: applied

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
xsalonx: I don't whethere there is better solution (still using sequelize)
1. This attribute must be calculated
2. so filtering by it must be done using subquery in WHERE clause or using HAVING clause
3. Using HAVING clause breaks counting, `Model.findAndCountAll` dosn't work anymore as well as `Model.count`. Simpler, `findAll` works fine, but it doesn't provide counting information.
4. So literal in where clause...

lib/database/adapters/LhcPeriodStatisticsAdapter.js
martinboulais: But I suppose you can use `.beamType` no?

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
martinboulais: I think you should not handle such a case in here.
Did you noticed RC about this run?

martinboulais (changes requested):

lib/database/adapters/LhcPeriodStatisticsAdapter.js
xsalonx: no

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
xsalonx: Well, I tried achieve the same what was in RCT and it is. If it was considered it's very old.
And if not handle it here, then I need to add some group concats, and introduce changes in API and front, which would bring meaningless information

xsalonx: If someone would need this info it might me added later, LHC Periods page is for general information and this is not one

lib/database/adapters/LhcPeriodStatisticsAdapter.js
xsalonx: as I said object destruction doesn't work, which is shorthand for what you wrote

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
xsalonx: changed, it doesn't apply anymore, as problematic run is of bad quality, see https://github.com/AliceO2Group/Bookkeeping/pull/1253

PR #1262, opened 2023-11-30
[O2B-1057] Use infoLoggerButtonGroup component in Environments and Runs overviews

SandroBoelsz:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- The infoLoggerButtonGroup itself can be used to navigate to a specific details page
- The infoLoggerButtonGroup dropdown can be used to copy the identifiable parameter
- The infoLoggerButtonGroup dropdown can be used to navigate to the infologgers
- The Environments overview page now uses the buttonGroup UX from the Runs overview page
- The old infologger column has been removed from the Environments overview page

- Removed displayEnvironmentInfologgerLinks component previously used in old infologger column
- Added an infoLoggerButtonGroup component which combines Runs and Environments InfoLoggerButtonGroup components
- Added a CopyToClipboardComponent which can be used to copy a given key and value to the clipboard
- Added an indetifierColumnAndActionsDropdown component which can be used to render a buttongroup with an identfier and a dropdown.

Changes made to the database:
- No changes to database


martinboulais (changes requested):

lib/public/components/common/selection/infoLoggerButtonGroup/infoLoggerButtonGroup.js
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @param {Object} parameters the parameters that specify what links to display for the details page and the infologgers.
+ * @return {Component} the infologger button group component
+ */
+export const infoLoggerButtonGroup = (parameters) => {
+    const { flp: flpInfologger, epn: epnInfologger } = getInfologgerLink(parameters);
+    const { environmentId, runNumber } = parameters;
+    const key = runNumber ? { runNumber } : { environmentId };
+    const value = runNumber || environmentId;
+    const page = runNumber ? 'run-detail' : 'env-details';
```
martinboulais, line 33: I think that here you are coupling a bit too much your component with specific pages.
I would say that you should have a component for example `identifierColumndAndActionsDropdown` that takes as a arguments the page, the link content (value for you), the parameters (not key, parameters), and a list of dropdown entries. 
In this component, you display the component you have here, and you build the dropdown options with the copy button and the list of given dropdown entries.

Then you will have two calls, one for run and one for envs.

Because here you are really adding too much logic in this components (info about runs, info about environment), making it quite hard to maintain.

martinboulais (changes requested):

lib/public/components/common/selection/infoLoggerButtonGroup/identifierColumnAndActionsDropdown.js
```diff
@@ -0,0 +1,41 @@
...
+import { iconCaretBottom } from '/js/src/icons.js';
+import { h } from '/js/src/index.js';
+
+/**
+ * Renders a button group with an identifier column and actions dropdown.
+ *
+ * @param {string} page The page to link to.
+ * @param {string} linkContent The content of the link.
+ * @param {Object} parameters The parameters for the link.
+ * @param {Array<Component>} dropdownEntries The component entries for the dropdown.
```
martinboulais, line 25: Prefer `Component[]` to `Array<Component>`

lib/public/components/common/selection/infoLoggerButtonGroup/infoLoggerButtonGroup.js
```diff
@@ -11,42 +11,47 @@
...
+import { identifierColumnAndActionsDropdown } from './identifierColumnAndActionsDropdown.js';
+
+/**
+ * Generates an infologger link component with the given infologger link and label.
+ *
+ * @param {string} infologger The infologger link.
+ * @param {string} label The label for the link.
+ * @return {Component|null} The infologger link component, or null if the infologger is not available.
+ */
+const generateInfologgerLink = (infologger, label) => infologger && h('a', { href: infologger, target: '_blank' }, label);
```
martinboulais, line 26: I think generate is a bit too much, `getInfologgerLink` may be enough? Or even `infologgerLink`

The parameter `infologger` seems a bit misleading, I would put `infologgerUrl`

lib/public/components/common/selection/infoLoggerButtonGroup/infoLoggerButtonGroup.js
```diff
@@ -11,42 +11,47 @@
...
  *
- * @param {Object} parameters the parameters that specify what links to display for the details page and the infologgers.
- * @return {Component} the infologger button group component
+ * @param {Object} parameters The parameters that specify what links to display for the details page and the infologgers.
+ * @return {Component} The infologger button group component.
  */
 export const infoLoggerButtonGroup = (parameters) => {
     const { flp: flpInfologger, epn: epnInfologger } = getInfologgerLink(parameters);
+    const flpComponent = generateInfologgerLink(flpInfologger, 'Infologger FLP');
+    const epnComponent = generateInfologgerLink(epnInfologger, 'Infologger EPN');
```
martinboulais, line 38: Here I think that naming can be improved: `flpComponent` do not describe what they actually are, I would keep `flpInfologger` or `flpInfologgerLink`, same for EPN
You can (and should) also rename `getInfologgerLink` to `getInfologgerLinksUrls`

martinboulais (changes requested):

martinboulais (approved):

PR #1264, opened 2023-11-30
[O2B-1062] GET API LOG should allow for exclusion filter

DKR2002:
#### I have a JIRA ticket
- [ ] PR labels are selected

Notable changes for users:
- The exclusion filter was only applied to the author filter.
- The exclusion filter sign is set to an exclamation mark (!). Call the exclusion filter for the author, by placing this exclusion filter sign in front of the author you want to exclude (with or without a space).

- Added API test to test the new feature.
- For now the exclusion sign is set to an exclamation mark (!). This could be changed in the file itself (lib/usecases/log/GetAllLogsUseCase.js)

Changes made to the database:
-


lib/usecases/log/GetAllLogsUseCase.js
```diff
@@ -29,6 +29,27 @@ const { ApiConfig } = require('../../config/index.js');
...
+ * removes the exclusion sign, if present.
+ *
+ * @param {string} dtoFilter the value to check for an exclusion sign.
+ * @returns {object} represents the parsed result.
+ * @property {boolean} excl indicates whether the first character of dtoFilter
+ *      is equal to exclSign.
+ * @property {string} strippedFilter the filtered string with the first
+ *      character removed if excl is true, otherwise, the original dtoFilter.
+ */
+const checkForFilterExclusion = (dtoFilter) => {
```
graduta, line 43: I would recommend that you add this method as a general method in a separate file so that other services can potentially re-use it

graduta (changes requested):

graduta (changes requested):
Hi @DKR2002 ,

PR looks almost complete but I believe you should tackle the corner case in which the user types `!` but nothing else. 
In this scenario, you should still return all logs which is not the case with your changes now. Moreover, make sure you add a test for it

graduta (approved):

PR #1268, opened 2023-12-01
[O2B-1058] lhc fill overview page should use button group on fill number column

galaxywolfv:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:

- Modify fillnumber column to display a buttongroup instead of just a number with the option to copy the identifier and have the options to use infologger.


lib/public/services/externalRouting/getInfologgerLinksUrls.js
```diff
@@ -19,14 +19,20 @@ const { CONFIGURATION: { FLP_INFOLOGGER_URL, EPN_INFOLOGGER_URL } } = window;
  * @param {object} filter the filter to apply to infologger
  * @param {string} [filter.environmentId] the id of environment to filter on (partition)
  * @param {number} [filter.runNumber] the runNumber of the run to filter on
+ * @param {Run[]} [filter.runs] the runs of the lhcFill to filter on
  * @return {{flp: (string|null), epn: (string|null)}} the infologger link URLs (null if no infologger is configured for the given type)
  */
-export const getInfologgerLinksUrls = ({ environmentId, runNumber }) => {
+export const getInfologgerLinksUrls = ({ environmentId, runNumber, runs }) => {
```
graduta, line 25: You should not add a new parameter which expects data similar to the other parameter `runNumber` especially as you are updating the same field `run` of the queryObject.

Instead, you should modify `runNumber` to expect an `array of numbers` instead of a simple `number` and update where the method is called with only one.

graduta (changes requested):

martinboulais (changes requested):

graduta (approved):

PR #1269, opened 2023-12-04
[O2B-1068] Accept and store EOR Reason from AliECS Core

CollinHva:
#### I have a JIRA ticket
- [ ] branch and/or PR name(s) include(s) JIRA ID
- [ ] PR labels are selected

Notable changes for users:
- From the AliECS Core, users can now update Eor reasons from a run with the help of gRPC

- RunService.updateEorReasonsOnRun now also accepts category and title as identifiers for eorReasonPatches

Changes made to the database:
- none

### DIFF
--- a/lib/server/controllers/gRPC/GRPCRunController.js
+++ b/lib/server/controllers/gRPC/GRPCRunController.js
@@ -43,13 +43,20 @@ class GRPCRunController {
 
     // eslint-disable-next-line require-jsdoc
     async Update(runPatchRequest) {
-        const { run, relations } = this.gRPCToRunAndRelations(runPatchRequest);
+        // Await conversion to keep correct sequence of code execution
+        const { run, relations } = await this.gRPCToRunAndRelations(runPatchRequest);
         const { runNumber } = run;
         delete run.runNumber;
 
         return this.runToGRPC(await this.runService.update(
             { runNumber },
-            { runPatch: run, relations: { runTypeName: relations.runType?.name } },
+            {
+                runPatch: run,
+                relations: {
+                    runTypeName: relations.runType?.name,
+                    eorReasons: relations.eorReasons,
+                },
+            },
         ));
     }
 
@@ -95,28 +102,27 @@ class GRPCRunController {
          * TODO: eorReasons is an empty array for some reason
          */
         if (gRPCRunRequest.eorReasons) {
-            // TODO: remove when fixed, this is temp EorReasons to go on with the task
+            /*
+             * TODO: remove when fixed, this is temp EorReasons to go on with the task
+             * No id means it is not yet in the db, so RunService.update will save as new one
+             */
             const tempEorReasons = [
                 {
-                    id: 1,
                     category: 'DETECTORS',
                     title: 'CPV',
                     description: 'test',
                 },
                 {
-                    id: 2,
                     category: 'DETECTORS',
                     title: 'TPC',
                     description: 'test',
                 },
                 {
-                    id: 7,
                     category: 'DETECTORS',
                     title: 'TPC',
                     description: 'test',
                 },
                 {
-                    id: null,
                     category: 'nonExisting',
                     title: 'TPC',
                     description: 'non existing reasonType',
@@ -125,17 +131,6 @@ class GRPCRunController {
 
             const convertedEorReasons = await convertEorReasonstoReasonTypeId(tempEorReasons);
             relations.eorReasons = convertedEorReasons;
-
-            //TODO: what to do with runId?
-
-            console.log(relations.eorReasons);
-
-            // TODO: conversion working, but not saving to db yet.
-
-            /*
-             * Logger.info('Getting here!');
-             * logger.info(gRPCRunRequest.eorReasons);
-             */
         }
 
         const { lhcPeriod } = gRPCRunRequest;
@@ -175,10 +170,12 @@ const convertEorReasonstoReasonTypeId = async (eorReasons) => {
             };
         }
 
-        return null; // Or handle the case when tempReasonType is not found
+        // Return null when tempReasonType is not found
+        return null;
     }));
 
-    return convertedEorReasons.filter(Boolean); // Remove null entries
+    // Remove null entries
+    return convertedEorReasons.filter(Boolean);
 };
 
 exports.GRPCRunController = GRPCRunController;
--- a/lib/server/controllers/gRPC/GRPCRunController.js
+++ b/lib/server/controllers/gRPC/GRPCRunController.js
@@ -12,6 +12,8 @@
  */
 
 const { runService } = require('../../services/run/RunService.js');
+const ReasonTypeRepository = require('../../../database/repositories/ReasonTypeRepository.js');
+const { QueryBuilder } = require('../../../database/utilities/QueryBuilder.js');
 
 /**
  * Controller to handle requests through gRPC RunService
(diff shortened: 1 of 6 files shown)


lib/server/services/run/RunService.js
```diff
@@ -276,13 +276,39 @@ class RunService {
...
-    const allReceivedReasonsExists = eorReasonPatches.every((eorReasonPatch) => reasonTypes.includes(eorReasonPatch.reasonTypeId));
+    // Determine the format of the eorReasonPatches (can be reasonTypeId || category and title)
+    if (eorReasonPatches[0]?.reasonTypeId) {
+        const reasonTypeIds = reasonTypes.map((reason) => reason.id);
+        const allReceivedReasonsExists = eorReasonPatches.every((eorReasonPatch) => reasonTypeIds.includes(eorReasonPatch.reasonTypeId));
 
-    if (!allReceivedReasonsExists) {
-        throw new Error('Provided reason types do not exist');
+        if (!allReceivedReasonsExists) {
+            throw new Error('Provided reason types do not exist');
```
graduta, line 287: Currently you are throwing an error if the EOR is not as expected.
While this is correct for the scenario in which the user selects manually the EOR, and we want to throw an error if the selection is bad, it is not the same for automated EOR set via the AliECS Core where we want to store the critical information such as `o2_time_stop`

To differentiate the 2 scenarios, you should split the grPCController method to make 2 calls to the runservice.
- one that updates the run with received information but EOR reasons. 
- one that updates the run with the recieved EOR reasons only. (this now can throw as we are not interested if it fails or not).
But the first scenario needs to always save the information, that is why we should remove EOR from the initial call.

graduta (dismissed):

lib/server/services/run/RunService.js
CollinHva: Should be fixed now

martinboulais (changes requested):

lib/server/controllers/gRPC/GRPCRunController.js
```diff
@@ -45,10 +45,17 @@ class GRPCRunController {
...
 
-        return this.runToGRPC(await this.runService.update(
+        // First call without Eor reason
+        this.runToGRPC(await this.runService.update(
             { runNumber },
             { runPatch: run, relations: { runTypeName: relations.runType?.name } },
         ));
+
+        // Second call only with Eor reasons
+        return this.runToGRPC(await this.runService.update(
```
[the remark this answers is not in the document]
CollinHva: For this, I added a try-catch in the controller as you mentioned. And I removed the throw from the RunService.updateEorReasonsOnRun function

lib/server/controllers/gRPC/GRPCRunController.js
```diff
@@ -41,14 +41,24 @@ class GRPCRunController {
 
     // eslint-disable-next-line require-jsdoc
     async Update(runPatchRequest) {
-        const { run, relations } = this.gRPCToRunAndRelations(runPatchRequest);
-        const { runNumber } = run;
-        delete run.runNumber;
+        try {
```
martinboulais, line 44: Here you don't need a try in the end?

Side note for you: try to avoid having huge try, because then you may have silent bugs. Only try the lines you expect to fail, not everything. The only exception is for example in the controller if you have a global catch to convert any error to a 500 response (for example)

lib/server/services/run/RunService.js
```diff
@@ -276,21 +276,38 @@ class RunService {
...
+        // Create a new map with all sub-maps
+        .flat());
+
+    const validPatches = [];
+    for (const patch of eorReasonPatches) {
+        // Use reasonTypeId if it exists, otherwise use category+title if it exists to get the reasonTypeId
+        const reasonTypeId = reasonTypesMap.get(patch.reasonTypeId ?? JSON.stringify([patch.category, patch.title])) ?? null;
+        // Add to the validPatches array if valid, if the patch is invalid ignore it
+        if (reasonTypeId !== null) {
+            // Set the reasonTypeId for adding to the db later
```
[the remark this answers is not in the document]
CollinHva: I have edited this in the comment

lib/server/services/run/RunService.js
```diff
@@ -276,21 +276,38 @@ class RunService {
...
+    /**
+     * @type {Map<number|string, number>}
+     */
+    const reasonTypesMap = new Map(reasonTypes
+        .map(({ id, category, title }) => [
+            // Set in the same map possibility to index by reasontypeId or by title/category
+            [id, id],
+            [JSON.stringify([category, title]), id],
+        ])
+        // Create a new map with all sub-maps
```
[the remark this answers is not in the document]
CollinHva: I have edited this in the comment

lib/server/controllers/gRPC/GRPCRunController.js
CollinHva: I think what you meant by this is because the RunService does not throw anymore I can leave the try behind. So, that is what I did, if this is not what you meant, please let me know and I will change it.

test/lib/server/services/run/RunService.test.js
```diff
@@ -310,4 +310,124 @@ module.exports = () => {
...
+                description: 'test2',
+            },
+        ];
+
+        const run = await runService.update(
+            { runNumber },
+            { relations: { eorReasons: eorReasons } },
+        );
+
+        expect(run.eorReasons).to.lengthOf(2);
```
martinboulais, line 333: Please also check the actual EoR reasons here, their ID & title/category for all tests

martinboulais (changes requested):

test/lib/server/services/run/RunService.test.js
CollinHva: I added this in the tests I made by checking the description, since I cannot check if the whole object is included in the run.eorReasons afterwards because it adds an id to the eorReason object in the tests and that makes that it is not equal anymore. That is why I compared the description because I made them unique in the tests.

test/lib/server/services/run/RunService.test.js
```diff
@@ -310,4 +310,137 @@ module.exports = () => {
...
+        // Description is unique, should be equal
+        for (let i = 0; i < run.eorReasons.length; i++) {
+            expect(run.eorReasons[i].description).to.equal(eorReasons[i].description);
```
martinboulais, line 336: Why don't you also test the reasonTypeId here?

test/lib/server/services/run/RunService.test.js
```diff
@@ -310,4 +310,137 @@ module.exports = () => {
...
```
martinboulais, line 364: Same thing here, you should be able to test the title, category and even that the id it the one expected (check in the seeders which reasonTypeId correspond to this title & category)

martinboulais (changes requested):

test/lib/server/services/run/RunService.test.js
CollinHva: Because the reasonTypeId is the same as the first object I am trying to add, the description is more unique here and that way you can be sure that it is the object I am referring to and that not the first object is added twice for example and the second one is not added. That is why I chose for the description, but if you would rather test the resaonTypeId than I can change it to that if you want.

test/lib/server/services/run/RunService.test.js
CollinHva: The title, category and id could be checked here, but the seeders are more hardcoded right, so is it not a bad idea to check for hardcoded seeders information for the reasonTypeId?

test/lib/server/services/run/RunService.test.js
CollinHva: Thank you for the explanation, I have implemented this now in the tests as you requested.

test/lib/server/services/run/RunService.test.js
CollinHva: Thank you for the explanation, I have implemented this now in the tests as you requested.

martinboulais (approved):

PR #1279, opened 2023-12-05
[O2B-952] Fixed a bug where the tally of replies was not accurately reflected on the log overview pages

SandroBoelsz:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- Replies are now tallied not only for the primary log but also for its associated child logs.
- The number of replies is now displayed on every log overview page, extending beyond just the log entries page.

- The reply count is now calculated using a custom recursive query, rather than within the repository itself.
- The reply count now encompasses both the replies on the root log and those on its child logs.
-  Addressed an SQL injection vulnerability in the existing getTagDistributionQuery of the LogRepository

Changes made to the database:
- None


lib/database/repositories/LogRepository.js
```diff
@@ -21,17 +21,41 @@ const { timestampToMysql } = require('../../server/utilities/timestampToMysql.js
...
+        WHERE parent_log_id = :parentLogId
+        UNION ALL
+        SELECT logs.id, logs.parent_log_id
+        FROM logs
+        JOIN reply_tree ON logs.parent_log_id = reply_tree.id
+    )
+    SELECT COUNT(*) AS reply_count FROM reply_tree;
+`, {
+    replacements: { parentLogId },
+});
```
martinboulais, line 58: The idea is here, nice job.
The main issue I see here is that you do one query per log entry, which is quite expensive if you have a lot of logs.
I propose to use instead:
```SQL
WITH RECURSIVE log_tree (originLogId, id) AS (
    SELECT id originId, id FROM logs -- In the anchor query, store the root log for which we want the tree
    WHERE id IN (7687, 7808) -- Pick the logs we are interested in, all at once
    UNION ALL
    SELECT originLogId, l.id FROM logs as l -- Then recurse, but always keep the origin to be able to group on it
        JOIN log_tree AS lt ON lt.id = l.parent_log_id
)
SELECT COUNT(*) - 1 FROM log_tree GROUP BY originLogId;
```

martinboulais (changes requested):

martinboulais (approved):

PR #1280, opened 2023-12-06
[O2B-1070] Add history column to environments overview page

saifrashed:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Not able to status message column anymore
- renamed Status to Current Status
- Added Status History which shows a sequence of coloured states

- A new format function is added that handles correct display of status history


lib/public/views/Environments/format/displayEnvironmentStatusHistory.js
```diff
@@ -0,0 +1,68 @@
...
+        MIX: 'M',
+        DESTROYED: 'DS',
+    };
+
+    /**
+     * Get correct color class from environmentname.
+     * @param {*} status One of the Environment statuses
+     * @returns {string} The classname used to color the acronoym.
+     */
+    const getColorClassEnvironment = (status) => {
```
graduta, line 39: You should not add again the status but look to complete `ColoredEnvironmentStatusComponent.js`

lib/public/views/Environments/format/displayEnvironmentStatusHistory.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+    };
+
+    const statusHistory = [];
+
+    // Iterate through historyItems
+    for (const { status } of historyItems) {
+        if (status in statusAcronyms) {
+            statusHistory.push({ status, content: statusAcronyms[status] });
+            statusHistory.push({ status: null, content: '-' });
```
graduta, line 58: If the purpose is simply to add a separator between the statuses, you should look at adding it in the display component below as `null` status is not a valid one.

graduta (changes requested):
First steps look good!
With a few more changes the PR will be ready to merge

lib/public/views/Environments/format/displayEnvironmentStatusHistory.js
saifrashed: - Added seperator in display component, used ternary operator to determine whether seperator should be shown. To avoid trailing seperator.

lib/public/views/Environments/format/displayEnvironmentStatusHistory.js
saifrashed: - ColoredEnvironmentStatusComponent has been completed to allow optional parameter called content. This will default to the status name, otherwise it takes given content.

graduta (approved):

PR #1281, opened 2023-12-06
[O2B-930] Add LHC period to run change notifier email

mennokonijn:
- [x] PR labels are selected

Notable changes for users:
- When updating the run quality, the user gets notified which LHC period belongs to that run

- none

Changes made to the database:
- none


lib/server/services/run/updateRun.js
```diff
@@ -111,6 +112,11 @@ exports.updateRun = async (identifier, payload, transaction) => {
         // Send notification if run quality has changed
         if (previousRun.runQuality !== runModel.runQuality) {
             const { runQualityChangeReason } = metadata;
+
+            const { lhcPeriodId } = runModel;
+            const lhcPeriod = await getOrCreateLhcPeriod(lhcPeriodId);
```
graduta, line 117: As you do not have any information on the name of the LHC Period, it is not a good idea here to create an lhc period just with the ID, as the ID is auto-incremented in the DB.

I would recommend to add a similar service for retrieving or failing the lhc period by id. (`getLhcPeriodOrFail.js`. You already have some examples getLogOrFail, etc.). Moreover, you already have the method for `findOne` in the Repository interface.

graduta: Moreover, if the period name cannot be found, then the user should not be impacted and the text parts for log entry for LHC Period should simply specify `UNKNOWN` for the lhc period

graduta (changes requested):

lib/server/services/run/updateRun.js
```diff
@@ -114,8 +114,8 @@ exports.updateRun = async (identifier, payload, transaction) => {
             const { runQualityChangeReason } = metadata;
 
             const { lhcPeriodId } = runModel;
-            const lhcPeriod = await getOrCreateLhcPeriod(lhcPeriodId);
-            const lhcPeriodName = lhcPeriod.name;
+            const lhcPeriod = await getLhcPeriodOrFail(lhcPeriodId);
```
graduta, line 117: As this method can throw, it means that your UpdateRun will fail as well due to this.
As we do not want that, you should either:
- use a `try/catch` in here and update the period name accordingly
- or, you should use the `getLhcPeriod` log which will return null instead of throwing error. 

I think you should go with second option as we are interested in continuing the log creation even if period name cannot be found.

PR #1293, opened 2023-12-12
[O2B-883]: Add bold line for detector quality in overview

mennokonijn:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Overview of runs now show quality by using a line under them in green/red

- none

Changes made to the database:
- none


lib/public/views/Runs/ActiveColumns/runsActiveColumns.js
```diff
@@ -78,9 +79,20 @@ export const runsActiveColumns = {
         visible: true,
         profiles: [profiles.none, 'lhcFill'],
         size: 'w-15 f6',
-        format: (detectors, run) => {
+        format: (_, run) => {
+            const detectors = run.detectorsQualities;
             if (detectors && detectors.length > 0) {
-                return [h('.badge.bg-gray-light.mh2.nDetectors-badge', run.nDetectors), `${detectors.toString()}`];
+                return [
```
graduta, line 85: You should look at extracting this formatting in an individual component so that it can be reused.
For example, we already provide a similar component `formatRunDetectors.js` which does a similar job for `run-details` page. 
Add a new component, named something like `formatRunDetectorsInline.js` (to suggest this will be a list of detectors inline)

graduta (changes requested):
In some cases, users will not be able to tell the difference between the qualities only through color and we should add a bit more information.
For example in `run-details` we use:
- `v` (check icon) + green to suggest good quality of detector
- `x` + red to suggest bad quality of detector

I would recommend in your case to think of something similar. Perhaps:
- `------` normal bold line (like you are doing now) + green for good quality
- ` - - - - ` visibly interrupted line  + red for bad quality

i would also recommend to increase a bit the weight of the line to be more visible

lib/public/views/Runs/format/formatRunDetectorsInline.js
```diff
@@ -0,0 +1,40 @@
...
+import { h } from '/js/src/index.js';
+import { RunDetectorQualities } from '../../../domain/enums/RunDetectorQualities.js';
+
+/**
+ * Format a list of detectors inline
+ *
+ * @param {Run} run the run for which detectors must be formatted
+ * @return {Component} the formatted run detectors inline
+ */
+export const formatRunDetectorsInline = (run) => {
```
graduta, line 23: I believe here you can simply expect only the `detectorQualities` and `nDetectors` instead of the whole run and update the documentation of the param to `RunDetectorQuality[]` from `Run` similar to `formatRunDetectors.js`

lib/public/views/Runs/format/formatRunDetectorsInline.js
```diff
@@ -0,0 +1,40 @@
...
+    const detectors = run.detectorsQualities;
+    if (detectors && detectors.length > 0) {
+        return [
+            h('.badge.bg-gray-light.mh2.nDetectors-badge', run.nDetectors), detectors.map((detector, index) => {
+                const isGoodQuality = detector.quality === RunDetectorQualities.GOOD;
+                const lineColor = isGoodQuality ? 'green' : 'red';
+                const lineStyle = isGoodQuality ? 'solid' : 'dashed';
+
+                return h('span', {}, [
+                    h('span', { style: { borderBottom: `4px ${lineStyle} ${lineColor}` } }, detector.name),
```
graduta, line 33: Instead of inline styling, i would recommend you define such classes in app.css and provide the element with className that it should use

graduta (changes requested):

lib/public/views/Runs/format/formatRunDetectorsInline.js
```diff
@@ -0,0 +1,39 @@
...
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { h } from '/js/src/index.js';
+import { RunDetectorQualities } from '../../../domain/enums/RunDetectorQualities.js';
+
+/**
+ * Format a list of detectors inline
```
graduta, line 18: Please provide more details with regards to what exactly `format` refers to in here

lib/public/views/Runs/format/formatRunDetectorsInline.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @return {Component} the formatted run detectors inline
```
graduta, line 20: `return` doc block should be added after the `param` list

test/public/lhcFills/detail.test.js
```diff
@@ -152,7 +152,7 @@ module.exports = () => {
 
         expect(runsTable.length).to.equal(4);
 
-        expect(await page.$eval('#runs tbody tr:nth-child(1) td:nth-child(2)', (element) => element.innerText)).to.equal('2FT0,ITS');
+        expect(await page.$eval('#runs tbody tr:nth-child(1) td:nth-child(2)', (element) => element.innerText)).to.equal('2FT0, ITS');
```
graduta, line 155: Space is limited for this particular table, that is why we were displaying detectors without a white space between the elements. Please revert this

lib/public/views/Runs/format/formatRunDetectorsInline.js
```diff
@@ -0,0 +1,39 @@
...
+ * @param {RunDetectorQuality[]} detectorsQualities The run detector quality per detector
+ * @param {number} nDetectors The number of run detectors
+ */
+export const formatRunDetectorsInline = (detectorsQualities, nDetectors) => {
+    if (detectorsQualities && detectorsQualities.length > 0) {
```
graduta, line 25:
```suggestion
    if (detectorsQualities?.length > 0) {
```
Browser supported by our applications are all available to use the optional chaining. [More](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

lib/public/views/Runs/format/formatRunDetectorsInline.js
```diff
@@ -0,0 +1,39 @@
...
+                const lineClass = isGoodQuality ? 'b-underline-good' : 'b-underline-bad';
+
+                return h('span', [
+                    h(`span.${lineClass}`, detector.name),
+                    index < detectorsQualities.length - 1 ? ', ' : '',
+                ]);
+            }),
+        ];
+    }
+    return '-';
```
graduta, line 38: I believe you can improve readbility of the code by first checking for bases cases such as detectorQualities length being 0 and returning `'-'`;
Then you can follow up with what to do if there are detectors present

lib/public/views/Runs/format/formatRunDetectorsInline.js
```diff
@@ -0,0 +1,39 @@
...
+ * Format a list of detectors inline
+ *
+ * @return {Component} the formatted run detectors inline
+ * @param {RunDetectorQuality[]} detectorsQualities The run detector quality per detector
+ * @param {number} nDetectors The number of run detectors
+ */
+export const formatRunDetectorsInline = (detectorsQualities, nDetectors) => {
+    if (detectorsQualities && detectorsQualities.length > 0) {
+        return [
+            h('.badge.bg-gray-light.mh2.nDetectors-badge', nDetectors), detectorsQualities.map((detector, index) => {
```
graduta, line 27: For code readability, you should put one element on each line.
In this particular case:
```suggestion
            h('.badge.bg-gray-light.mh2.nDetectors-badge', nDetectors),
            detectorsQualities.map((detector, index) => {
```

graduta, line 27: Also, I would recommend you extract to another component the design for each detector (name and underline) in the same file as a component that is not to be exported.

lib/public/views/Runs/format/formatRunDetectorsInline.js
```diff
@@ -0,0 +1,39 @@
...
+                const isGoodQuality = detector.quality === RunDetectorQualities.GOOD;
+                const lineClass = isGoodQuality ? 'b-underline-good' : 'b-underline-bad';
+
+                return h('span', [
```
graduta, line 31: There is no need to return a `span` within another `span` as you are coloring only the line underneath. 
You can simply return an array of these elements and use `flatMap` on line 27 when you call this

graduta (changes requested):

graduta (approved):

graduta (approved):

PR #1294, opened 2023-12-12
[1084] add getOrCreate function to UserService

CollinHva:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- none

- UserService now has a getOrCreate function

Changes made to the database:
- none


graduta (changes requested):
Hi @CollinHva ,

Please have a look at `CreateOrUpdateUser.js` usecase.
Usecases, are an old part of the codebase which we are actively migrating to the new `high/low` level services structure.

For this ticket, you should also look first at that functionality and migrate it to what you have implemented in this PR. This includes:
- the usecase itself
- tests for it
- usage of the usecase across the codebase

lib/server/services/user/createOrUpdateUser.js
```diff
@@ -0,0 +1,57 @@
...
+
+    // Check null OR undefined because !personid returns true with personid == 0
+    if (personid === null || personid === undefined) {
+        throw new BadParameterError('Cannot create without personid');
+    }
+
+    let user;
+    // GetUser expects an identifier with two options, so surround it with {}
+    const existingUser = await getUser({ externalUserId: personid });
+    if (existingUser) {
```
graduta, line 39: I believe you missplaced some functionality when migrating from the usecase.
I see in the old usecase, that we only update in the database if the retrieved values are different than the current one. This is done so that we do not make a new request to add same data.

Is there a reason you decided to skip that check?

graduta (changes requested):

test/lib/server/services/user/UserService.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+        const user = await userService.getOrCreate({ externalUserId: externalId }, name);
+
+        // User should exist now
+        expect(user).not.to.be.null;
+        // Id should be filled automatically
+        expect(user.id).not.to.be.null;
+        // ExternalId and name should be equal
+        expect(user.externalId).to.equal(externalId);
+        expect(user.name).to.equal(name);
```
martinboulais, line 35: Here this seems to me to be a bit of cyclic testing. Don't you have seeders for users?
If you don't do not bother it's still ok

lib/server/services/user/createOrUpdateUser.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    // Set user.access for server startup if present (WebUiServer)
+    if (access) {
+        user.access = access;
+    }
+    // Set user.username for server startup if present (WebUiServer)
+    if (username) {
+        user.username = username;
+    }
```
martinboulais, line 53: If these info are not persisted in the database, for me they should not go further than the high level service.

martinboulais (changes requested):

lib/server/services/user/createOrUpdateUser.js
CollinHva: Should be fixed now.

test/lib/server/services/user/UserService.test.js
CollinHva: I believe I do not have seeders. I try to make the tests so that they create data independently, so that they can run individually, like you adviced me to do because some tests are dependent on each other and that makes creating new tests a bit of a struggle sometimes, for example, for runs.
So, that is the main reason why I create the data in the same test.

lib/server/services/user/createOrUpdateUser.js
CollinHva: I am not sure what you mean by this, could you explain what you expect me to change here a bit differently maybe?

lib/server/services/user/UserService.js
```diff
@@ -11,8 +11,39 @@
...
+/**
+ * Global service to handle runs instances
+ */
+class UserService {
+    /**
+     * Find a user from the database or create it as a new user if it does not exist yet.
+     *
+     * @param {UserIdentifier} identifier the identifier of the user to find
+     * @param {string|null} name - username of the user
+     * @returns {User} - found user | newly created user
```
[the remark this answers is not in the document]
CollinHva: You are right, I did not know that that was needed, I implemented it now, thanks for the tip!

lib/server/services/user/UserService.js
```diff
@@ -11,8 +11,39 @@
...
+const { getUser } = require('./getUser.js');
+
 /**
  * @typedef UserIdentifier object to uniquely identify a user
  * @property {number} [userId] the id of the user in the application's context
  * @property {number} [externalUserId] the CERN user's id
  */
+
+/**
+ * Global service to handle runs instances
```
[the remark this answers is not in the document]
CollinHva: Done

lib/server/services/user/createOrUpdateUser.js
CollinHva: I think I have found a solution to this, see WebUiServer.js for my implemented solution. The only part this is used is in the call from WebUiServer.js, but those variables are also present in request.session, so I can just move this to WebUiServer.js because no other code is using or going to use that.

martinboulais (approved):

graduta (approved):

PR #1296, opened 2023-12-12
[O2B-1066]: Added extra check for empty for log environment filter

mennokonijn:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Error does not show up when going back to no value

- none

Changes made to the database:
- none


graduta (changes requested):
Hi @mennokonijn ,

Instead of adding an extra check, I would recommend you having a look a bit below at line 250 to see how the LHC filter behaves and removes false positives such as empty strings

graduta (approved):

PR #1298, opened 2023-12-13
[O2B-1083] Fix a bug in LHC overview pages, now displaying missing dates as '-' instead of the default epoch time

SandroBoelsz:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- Missing dates on the LHC overview pages are now shown to the user as '-' instead of 01/01/1970 01:00:00

- The LHCFillAdapter now verifies whether a date is null/undefined and refrains from using new Date() when encountering missing values

Changes made to the database:
- None

### DIFF
--- a/lib/database/adapters/LhcFillAdapter.js
+++ b/lib/database/adapters/LhcFillAdapter.js
@@ -49,9 +49,9 @@ class LhcFillAdapter {
 
         return {
             fillNumber,
-            stableBeamsStart: new Date(stableBeamsStart).getTime(),
-            stableBeamsEnd: new Date(stableBeamsEnd).getTime(),
-            stableBeamsDuration,
+            stableBeamsStart: stableBeamsStart ? new Date(stableBeamsStart).getTime() : stableBeamsStart,
+            stableBeamsEnd: stableBeamsEnd ? new Date(stableBeamsEnd).getTime() : stableBeamsEnd,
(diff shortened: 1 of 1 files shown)


lib/database/adapters/LhcFillAdapter.js
```diff
@@ -49,9 +49,9 @@ class LhcFillAdapter {
 
         return {
             fillNumber,
-            stableBeamsStart: new Date(stableBeamsStart).getTime(),
-            stableBeamsEnd: new Date(stableBeamsEnd).getTime(),
-            stableBeamsDuration,
+            stableBeamsStart: stableBeamsStart ? new Date(stableBeamsStart).getTime() : stableBeamsStart,
+            stableBeamsEnd: stableBeamsEnd ? new Date(stableBeamsEnd).getTime() : stableBeamsEnd,
+            stableBeamsDuration: stableBeamsDuration ? new Date(stableBeamsDuration).getTime() : stableBeamsDuration,
```
martinboulais, line 54: I think that the duration is not a Date

martinboulais (changes requested):

lib/database/adapters/LhcFillAdapter.js
SandroBoelsz: you're right, it looks like copilot went along and changed that field as well. My mistake, I fixed it now. Sorry.

martinboulais (approved):

PR #1303, opened 2023-12-14
[O2B-926] Added button to overview for PHYSICS filter

mennokonijn:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- extra button in runs overview to filter on PHYSICS

-

Changes made to the database:
-


lib/public/views/Runs/Overview/RunsOverviewPage.js
```diff
@@ -57,3 +59,23 @@ const exportRunsButton = ({ runsOverviewModel, modalModel }) => h('button.btn.bt
...
+/**
+ * Builds a button which will toggle the physics filter
+ * @param {OverviewModel} runsOverviewModel the model of the runs overview
+ * @returns {vnode} with button
+ */
+const showPhysicsButton = (runsOverviewModel) => {
+    const isPhysicsActive = runsOverviewModel.isDefinitionInFilter(RunDefinition.Physics);
+    const buttonColor = isPhysicsActive ? 'primary' : 'secondary';
+
+    return h(`button.btn.btn-${buttonColor}.w-10.h2.ml2#show-physics-trigger`, {
```
martinboulais, line 72: Avoid as much as possible to put fixed width/height on your components, especially on components that have a fixed size (such as a button).
Same for margin. When you just want spacing between items, please use gap (we have classes for it, it's `g1`, `g2` and `g3`) on the flex parent. Doing so, please also remove the padding of the run number input

lib/public/domain/enums/RunDefinition.js
```diff
@@ -19,3 +19,5 @@ export const RunDefinition = {
     Calibration: 'CALIBRATION',
     Commissioning: 'COMMISSIONING',
 };
+
+export const RUN_DEFINITIONS = Object.values(RunDefinition);
```
martinboulais, line 23: Nice one 👍🏻

martinboulais (changes requested):

martinboulais:
Please also rename the PR to use the proper format

docs/get-started.md
```diff
@@ -26,7 +26,7 @@ Use the npm script (the `docker-compose` command must be available to npm)
 npm run docker-run
 ```
 
-You will then see in this tabs the server's log
+You will then see in these tabs the server's log
```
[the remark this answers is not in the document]
mennokonijn: I got this as an automatic fix from my code editor. And I also checked if this change was correct. Because tabs is plural this needs to be these instead of this, so that is why I changed it.

lib/public/views/Runs/Overview/RunsOverviewPage.js
mennokonijn: Fixed this issue, the only problem now is a layout shift when the text changes to show all, so there are a few solutions to this.

1. We only show the text Show PHYSICS and just make it blue when it is active and grey when not
2. We do use the change in text and thus use a fixed with for the button
3. We make it a checkbox

I let the decision be yours, for now I did not switch the text and this chose for option 1

docs/get-started.md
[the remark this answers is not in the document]
martinboulais: I read too quickly indeed, my bad

graduta (approved):

graduta (approved):

PR #1304, opened 2023-12-15
[O2B-1085] update proto files and grpc run controller with new user struct

CollinHva:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- none

- common.proto User message now has optional id
- common.proto Run message now has optional User user_o2_start and user_o2_stop variables
- common.proto RunCreationRequest and RunUpdateRequest message now have optional User user_o2_start and user_o2_stop variables

Changes made to the database:
- none


lib/server/controllers/gRPC/GRPCRunController.js
```diff
@@ -34,13 +36,23 @@ class GRPCRunController {
 
     // eslint-disable-next-line require-jsdoc
     async Create(newRunRequest) {
+        // Run can only be updated or created with a valid user
+        if (!await this.userService.validateUserFromGRPCRunRequest(newRunRequest.user_o2_start, newRunRequest.user_o2_stop)) {
```
martinboulais, line 40: The services have no knowledge of API. gRPC is API, it should not be in services.

The tickets has a mistype, you should not return if there is no user, but simply not update this field

martinboulais (changes requested):

lib/server/controllers/gRPC/GRPCRunController.js
CollinHva: Should be fixed now

graduta (approved):

PR #1307, opened 2023-12-15
[O2B-1095] remove attachment token dependency

xsalonx:
#### I DON'T have JIRA ticket
- [x] tests are added

Notable changes for users:
- fix attachment downloading

- refactor logs component
- adds missing test for attachment downloading

Changes made to the database:
- NA

### DIFF
--- a/lib/public/components/Log/log.js
+++ b/lib/public/components/Log/log.js
@@ -82,7 +82,7 @@ const fieldFormatting = {
                                 a.click();
                                 a.remove();
                             }),
-                    class: 'break-all' },
+                    class: 'break-all clickable' },
             }),
             { bypassRouter: true },
         ),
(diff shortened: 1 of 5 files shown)


martinboulais:
The idea is good. However, what happens if you have a huge file (which for now is not limited)?
The user will see nothing happening (no remote data), there is no chunk-downloading and the whole file will be put in memory?

lib/public/components/Log/log.js
```diff
@@ -72,13 +70,24 @@ const logFields = (authenticationToken) => ({
             attachments,
             ({ fileName, originalName, id }) => ({
                 content: originalName,
-                href: `/api/attachments/${id}?token=${authenticationToken}`,
-                attributes: { download: fileName, class: 'break-all' },
+                attributes: {
+                    onclick: async () =>
+                        fetchClient(`/api/attachments/${id}`)
+                            .then((res) => res.blob())
+                            .then((blob) => URL.createObjectURL(blob))
```
martinboulais, line 77: When you create an object URL you should release it.

martinboulais (changes requested):

PR #1315, opened 2023-12-22
[1086] Store user starting/stopping a run in database

CollinHva:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- none

- RunService.update and create functions now also accept user_o2_start and user_o2_stop.

Changes made to the database:
- Run entity now has optional user_o2_start and user_o2_stop, both of type Integer.


lib/database/migrations/20231215113344-add-stop-and-start-users-to-run.js
```diff
@@ -0,0 +1,22 @@
...
+        await queryInterface.addColumn('runs', 'user_o2_start', {
+            allowNull: true,
+            type: Sequelize.INTEGER,
+            default: null,
+        });
+        await queryInterface.addColumn('runs', 'user_o2_stop', {
+            allowNull: true,
+            type: Sequelize.INTEGER,
+            default: null,
+        });
```
martinboulais, line 15: Please add the foreign key

lib/server/services/run/RunService.js
```diff
@@ -194,6 +207,17 @@ class RunService {
...
+        if (user_o2_start?.externalId) {
+            // Create the user if it does not yet exist
+            const user = await userService.getOrCreate({ externalUserId: user_o2_start.externalId }, user_o2_start.name);
+            runPatch.user_o2_start = user.id;
+        }
+        if (user_o2_stop?.externalId) {
+            // Create the user if it does not yet exist
+            const user = await userService.getOrCreate({ externalUserId: user_o2_stop.externalId }, user_o2_stop.name);
+            runPatch.user_o2_stop = user.id;
+        }
```
martinboulais, line 219: This should be factorized

proto/common.proto
```diff
@@ -104,8 +104,9 @@ message Run {
   // A string that marks the period of the lhc
   string lhcPeriod = 31;
   // Optional users that stopped or started the run (can also be undefined)
-  optional User user_o2_start = 32;
-  optional User user_o2_stop = 33;
+  // id in this case because Run has a user id in bookkeeping
+  optional int32 user_o2_start = 32;
+  optional int32 user_o2_stop = 33;
```
martinboulais, line 109: camelCase

lib/server/services/run/RunService.js
```diff
@@ -145,6 +146,17 @@ class RunService {
             runType = await getOrCreateRunType({ name: runTypeName });
         }
 
+        if (user_o2_start?.externalId) {
```
martinboulais, line 149: user_o2_start and stop are numbers no? Where does the externalId comes from?!

martinboulais (changes requested):

lib/database/adapters/RunAdapter.js
```diff
@@ -179,6 +181,8 @@ class RunAdapter {
             otherFileCount,
             otherFileSize: otherFileSize !== null ? String(otherFileSize) : null,
             lhcFill,
+            user_o2_start,
+            user_o2_stop,
```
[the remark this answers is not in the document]
CollinHva: done

lib/database/adapters/RunAdapter.js
```diff
@@ -265,6 +269,8 @@ class RunAdapter {
             concatenatedDetectors: entityObject.detectors,
             definition: entityObject.definition,
             calibrationStatus: entityObject.calibrationStatus,
+            user_o2_start: entityObject.user_o2_start,
+            user_o2_stop: entityObject.user_o2_stop,
```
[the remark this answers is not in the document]
CollinHva: done

lib/database/models/run.js
```diff
@@ -204,6 +204,17 @@ module.exports = (sequelize) => {
...
+        user_o2_start: {
+            allowNull: true,
+            type: Sequelize.INTEGER,
+            default: null,
+        },
+        user_o2_stop: {
+            allowNull: true,
+            type: Sequelize.INTEGER,
+            default: null,
+        },
```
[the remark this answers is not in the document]
CollinHva: I changed it to what you said: with id, because it is indeed a fk to the id

lib/domain/dtos/UpdateRunDto.js
```diff
@@ -39,6 +39,8 @@ const BodyDto = Joi.object({
     detectorsQualitiesChangeReason: Joi.string().optional(),
     calibrationStatus: Joi.string().valid(...RUN_CALIBRATION_STATUS).optional(),
     calibrationStatusChangeReason: Joi.string().optional(),
+    user_o2_start: Joi.number().optional(),
+    user_o2_stop: Joi.number().optional(),
```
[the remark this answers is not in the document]
CollinHva: done

lib/domain/entities/Run.js
```diff
@@ -73,6 +73,8 @@
  * @property {EorReason[]} eorReasons
  * @property {RunType|null} runType
  * @property {LhcFill} lhcFill
+ * @property {number|null} user_o2_start
+ * @property {number|null} user_o2_stop
```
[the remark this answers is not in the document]
CollinHva: If I did not misinterpret this comment, I implemented the new relation as you said and added a comment to make it clearer. It is indeed a relation, so it is in the last part like it is now right?

lib/server/services/run/RunService.js
CollinHva: Because the gRPC run service sends a User message and not an id, because the external AliECS Core does not know the user id in the bookkeeping db, so the gRPC run service sends a User without id, but with an externalId. And with the externalId I made a function getOrCreate to get the user from the bookkeeping db. After doing that function, I have the full user available, now including the id, so that I can use that for further processing (the run model can now use this user id acquired in the described functionality).

This is described in the parent ticket: "The protofile should be updated to include a new struct defined in common.proto which defines a user containing: number external_id, string name (which is actually a username)."

lib/database/migrations/20231215113344-add-stop-and-start-users-to-run.js
CollinHva: done

lib/server/services/run/RunService.js
CollinHva: done

proto/common.proto
CollinHva: done

lib/database/adapters/RunAdapter.js
```diff
@@ -125,6 +125,8 @@ class RunAdapter {
             logs,
             definition,
             calibrationStatus,
+            userIdO2Start,
+            userIdO2Stop,
```
martinboulais, line 129: Could you just put these after timeTrgEnd? Same for everywhere they all appear

lib/database/migrations/20231215113344-add-stop-and-start-users-to-run.js
```diff
@@ -0,0 +1,45 @@
...
+        await queryInterface.addConstraint('runs', {
+            fields: ['user_id_o2_stop'],
+            type: 'foreign key',
+            name: 'fk_user_o2_stop',
+            references: {
+                table: 'users',
+                field: 'id',
+            },
+        });
```
martinboulais, line 34: It's not critical, but you can put the FK in the column creation

lib/database/migrations/20231215113344-add-stop-and-start-users-to-run.js
```diff
@@ -0,0 +1,45 @@
...
+    }),
+
+    down: async (queryInterface) => queryInterface.sequelize.transaction(async () => {
+        await queryInterface.removeColumn('runs', 'user_id_o2_start');
+        await queryInterface.removeColumn('runs', 'user_id_o2_stop');
+
+        // Delete fk constraints
+        await queryInterface.removeConstraint('runs', 'fk_user_o2_start');
```
martinboulais, line 42: You need to delete the fk before deleting the column

lib/domain/entities/Run.js
```diff
@@ -73,6 +73,8 @@
  * @property {EorReason[]} eorReasons
  * @property {RunType|null} runType
  * @property {LhcFill} lhcFill
+ * @property {number|null} userIdO2Start relation to the user id that started the run
+ * @property {number|null} userIdO2Stop relation to the user id that stopped the run
```
martinboulais, line 77: Please do the same in `SequelizeRun`

lib/domain/entities/Run.js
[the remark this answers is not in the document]
martinboulais: What I mean by relation is, in a property, the related model. The relation is not the fk: 
For example for `lhc fill` you have:
`fillNumber` which is the FK (`number`)
`lhcFill` which is the actual related entity (`LhcFill` instance)
In your case, you only have the foreign key, which should be with the columns.

proto/common.proto
```diff
@@ -104,8 +104,9 @@ message Run {
   // A string that marks the period of the lhc
   string lhcPeriod = 31;
   // Optional users that stopped or started the run (can also be undefined)
-  optional User user_o2_start = 32;
-  optional User user_o2_stop = 33;
+  // id in this case because Run has a user id in bookkeeping
+  optional int32 userO2Start = 32;
+  optional int32 userO2Stop = 33;
```
martinboulais, line 109: I think this is not useful, because it will contain the user id, which is specific to bkp. 
I would simply omit this field from gRPC, and we'll add it when one of the client needs it.
(Only for this, not in the run service!)

lib/server/services/run/RunService.js
martinboulais: Ok you are right

lib/domain/dtos/UpdateRunDto.js
```diff
@@ -39,6 +39,8 @@ const BodyDto = Joi.object({
     detectorsQualitiesChangeReason: Joi.string().optional(),
     calibrationStatus: Joi.string().valid(...RUN_CALIBRATION_STATUS).optional(),
     calibrationStatusChangeReason: Joi.string().optional(),
+    userIdO2Start: Joi.number().optional(),
+    userIdO2Stop: Joi.number().optional(),
```
martinboulais, line 43: Don't bother with this API, you can leave it without it I suppose.

lib/server/services/run/RunService.js
```diff
@@ -318,6 +328,23 @@ const updateEorReasonsOnRun = async (runId, userName, eorReasonPatches) => {
...
 
+/**
+ * Method to process a user for the run
+ *
+ * @param {User|null} userData - user object, either userO2Start or userO2Stop
+ * @param {number} [userData.externalId] - external identifier of the user
+ * @param {string|null} [userData.name] - name of the user or empty
+ * @returns {Promise<number|null>} - returns the id of the user or null if the userData is null
+ */
+const processUser = async (userData) => {
```
martinboulais, line 339: I would rather name this function `getUserIdFromExternalId` and take as parameter the external id directly.

martinboulais (changes requested):

lib/server/services/run/RunService.js
CollinHva: done

proto/common.proto
CollinHva: So, do you want me to change it to userIdO2Start then? And I already made functionality for this, so maybe it is good to keep it for now, since you expect it to maybe be needed by a client in the future?

lib/database/adapters/RunAdapter.js
CollinHva: done

lib/domain/entities/Run.js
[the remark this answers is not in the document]
CollinHva: Ah ok, I put it after timeTrgEnd now like you mentioned in the next comment, so now it is in the columns section

lib/database/migrations/20231215113344-add-stop-and-start-users-to-run.js
CollinHva: fk constraint is now in the columns, so no need to delete them anymore, which fixes this problem to

lib/domain/entities/Run.js
CollinHva: done

lib/domain/dtos/UpdateRunDto.js
CollinHva: done

lib/server/services/run/RunService.js
martinboulais: I apologize, I read to quickly.
I would rather name it getOrCreateUser, take the userData as you did, and return the id or null.
The you simply have something like
```js
newRun.userIdO2Start = (await getOrCreateUser(userData))?.id
```

proto/common.proto
martinboulais: No I prefer to simply delete it for now. If at some point we need it, it will not be hard to remove it.
Also: In this proto, you added `id` in User message (in a previous PR https://github.com/AliceO2Group/Bookkeeping/commit/a9641894a0b68e1679a871a02e8317a458e8f956#diff-1747d3070a2311a07389bc9e83309513fbf9f16f7aac7a80988344c38fb98635R127). The thing is, `id` is Bookkeeping internal, and it should not be shared through APIs. I suppose it is not used, and it should be removed.

I understand it is tempting to add code that may be used later, but in the end you have a big codebase with code not used, and the bigger the codebase is, the harder it is to maintain.

martinboulais (changes requested):

proto/common.proto
CollinHva: George told me to make the id optional, that is the only thing I changed, because it is used for log, see the message from George below:
"Indeed, I forgot that we have it for creating a log. In this scenario, the id might be null so you use the existing user message but make the ID optional".

So, I think it is used, I only made it optional for this specific task.

I removed the User fields in common.proto now, but I kept the User message as it is, is that good or do you still want me to remove it with the above information George told me?

Thank you for the explanation, I understand now that the codebase will get increasingly complex.

lib/server/services/run/RunService.js
CollinHva: No problem, I changed it now.

lib/server/services/run/RunService.js
```diff
@@ -331,14 +331,15 @@ const updateEorReasonsOnRun = async (runId, userName, eorReasonPatches) => {
 /**
  * Method to process a user for the run
  *
- * @param {number} externalId - external identifier of the user
- * @param {string|null} name - name of the user or empty
+ * @param {User|null} userData - user object, either userO2Start or userO2Stop
```
martinboulais, line 334: It is not really a user, but a user identification/creation paylaod. You should type it as `object|null` then specify its properties (as you already done)

martinboulais (changes requested):

lib/server/services/run/RunService.js
CollinHva: I understand, I changed it now

proto/common.proto
martinboulais: I do not see where it can be used, but you are right, let's keep it from the `User` (it was my fault to put it here in the first place)

martinboulais (approved):

PR #1319, opened 2024-01-08
[O2B-1093] abstract http synchronizer

xsalonx:
### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- NA

- add abstraction of data fetching from external services 

Changes made to the database:
- NA


lib/server/monalisaSynchronizers/AbstractHTTPSynchronizer.js
```diff
@@ -0,0 +1,88 @@
...
+
+const { Log } = require('@aliceo2/web-ui');
+
+/**
+ * AbstractHTTPSynchronizer
+ * The class provides schema for excecuting process of data synchronization with external service
+ * Its behaviour can be customized with overriding abstract methods
+ * Communication is carried via http with json data format
+ */
+class AbstractHTTPSynchronizer {
```
martinboulais, line 24: I don't understand the point of this abstraction, maybe it would make more sense if I see an implementation.
What is the advantage of having a class that does all of this?
If I rephrased, what is the purpose of this class? It has a lot of reason to change (if the API changes, to gRPC for example, if we change the data storage and so on).

xsalonx: There is need to fetch data from three endpoints of MonALISA. Each process of data fetching looks more or less the same P: (fetch data) -> (adjust data) -> (update DB).
- P1 - fetching data passes
- P2 - fetching data passes details - need to be done alongside with P1
- P3 - fetching Monte Carlo productions (Simulation passes)

Due to this there is an abstract class which will be inherited by three other.
-----
If API change and we have no such a abstraction I think there will be more changes required.
Storage? What about Sequelize?

lib/server/monalisaSynchronizers/AbstractHTTPSynchronizer.js
```diff
@@ -0,0 +1,88 @@
...
+     * @param {object} _rawResponse raw data acquired from external service
+     * @return {*} adjusted data
+     */
+    async processRawResponse(_rawResponse) {
+        throw new Error('Abstract function call');
+    }
+
+    /**
+     * Implements logic for inserting/updating database
+     * @abstractMethod
```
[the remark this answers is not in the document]
xsalonx: done

lib/server/monalisaSynchronizers/AbstractHTTPSynchronizer.js
```diff
@@ -0,0 +1,88 @@
...
+            .then(() => true) // Passed without major errors
+            .catch(async (fatalError) => {
+                this.logger.error(`Synchronizer fatal error: ${fatalError.message}`);
+                return false; // Major error occurred
+            });
+    }
+
+    /**
+     * Used to process raw response to custom format, remove unnecessary data
+     * @abstractMethod
```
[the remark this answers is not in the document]
xsalonx: done

lib/server/monalisaSynchronizers/AbstractHTTPSynchronizer.js
```diff
@@ -0,0 +1,88 @@
...
+     * Combine logic of fetching data from service like MonALISA, processing and inserting to database
+     * @param {URL} url endpoint to fetch data from
+     * @param {object} context db action context
+     * Besides the given arguemnt the method depends on two abstract methods to be overriden
+     * @see AbstractHTTPSynchronizer.processRawResponse
+     * @see AbstractHTTPSynchronizer.executeDbAction
+     *
+     * @returns {boolean} - true if process was finalized without major errors, otherwise false,
+     * Major errors are understood as ones indicating that further synchronization is purposeless: e.g. due to networ error, invalid certificate.
+     * Minor errors are understood as e.g. managable ambiguities in data.
```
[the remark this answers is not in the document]
xsalonx: Previously I implemented other class in a way that it was more convenient to have something like that here, but, anyway I reviewed it and, indeed, it's not necessary so I removed it

lib/server/monalisaSynchronizers/AbstractHTTPSynchronizer.js
xsalonx: For P1 and P2 it is like
for each DB update of P1 do entire P2 within one transaction - what is to update Data Pass and its details together in order to keep data consistent

martinboulais: Please put at least one concrete class first, then when you bring the second you may bring the abstraction.
Here, I see no really advantages of this abstraction except a lot of configurable code that will always be really specific.
Please, put the first concrete class in a first PR, then the abstraction with a second concrete class when it makes sense. Or, if you use the abstraction somewhere independently of the concrete classes I'm also ok.

martinboulais: Here you are mixing two low level concepts: database, and API. If you want to have such a thing, it should be very generic, for example `ExternalDataSynchronizer`, to which you provide a way to fetch data from API, and a way to write data to storage, with adapters on both sides... But it's probably too much work for nothing.

PR #1323, opened 2024-01-11
[O2B-1032] improvements to the environment table

mennokonijn:
- [x] PR labels are selected

Notable changes for users:
-

- Change made to the balloon, this now also checks if the height overflows

Changes made to the database:
-


lib/public/components/common/popover/overflowBalloon.js
```diff
@@ -23,7 +23,8 @@ import { PopoverAnchors } from './PopoverEngine.js';
  * @param {HTMLElement} element the element to look for overflow
  * @returns {boolean} true if the element overflow
  */
-const doesElementOverflow = (element) => Boolean(element) && element.offsetWidth < element.scrollWidth;
+const doesElementOverflow = (element) =>
+    Boolean(element) && element.offsetWidth < element.scrollWidth || Boolean(element) && element.offsetHeight < element.scrollHeight;
```
martinboulais, line 27: Please don't put newline after `=>`. In this case, I think the code would be more clear with a body and intermediate variables `overvlowVertically` and `overflowHorizontally`

lib/public/app.css
```diff
@@ -162,6 +162,10 @@ th.text-center, td.text-center {
     text-align: center;
 }
 
+.table-cl-mh-2 {
+    max-height: 48px;
+}
+
```
martinboulais, line 168: You should avoid as much as possible to use fixed pixels height. I would rather suggest to use `line-clamp` which seems to be widely available (with `-webkit-` prefix) https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp#truncating_a_paragraph

lib/public/views/Environments/format/displayEnvironmentStatusHistory.js
```diff
@@ -28,10 +28,20 @@ export const displayEnvironmentStatusHistory = (environment) => {
         .map(({ status }) => ({ status, content: STATUS_ACRONYMS[status] }));
 
     return h(
-        '.flex-row',
-        statusHistory.map((value, index) => h('.flex-row', [
-            coloredEnvironmentStatusComponent(value.status, value.content),
-            h('', index < statusHistory.length - 1 ? '-' : ''),
-        ])),
+        '.flex-row.flex-wrap.table-cl-mh-2',
+        [statusHistory.map((value, index) => getStatusHistoryItem(value, index, statusHistory.length))],
```
martinboulais, line 32:
```suggestion
        statusHistory.flatMap((value) => [
            coloredEnvironmentStatusComponent(value.status, value.content),
            h('', '-')
        ]).slice(0, -1),
```

lib/public/views/Environments/ActiveColumns/environmentsActiveColumns.js
```diff
@@ -29,21 +29,30 @@ export const environmentsActiveColumns = {
...
+    },
     updatedAt: {
         name: 'Updated At',
         visible: true,
         sortable: false,
         title: true,
-        size: 'w-15',
-        format: (timestamp) => formatTimestamp(timestamp),
+        size: 'w-10',
+        format: (_, environment) => formatEnvironmentDates(environment.updatedAt),
```
martinboulais, line 47: Why replacing `timestamp` by `(_, environemnt)` to in the end get the same property?

lib/public/views/Environments/ActiveColumns/environmentsActiveColumns.js
```diff
@@ -29,21 +29,30 @@ export const environmentsActiveColumns = {
...
     },
     createdAt: {
         name: 'Created At',
         visible: true,
         sortable: false,
         title: true,
-        size: 'w-15',
-        format: (timestamp) => formatTimestamp(timestamp),
+        size: 'w-10',
+        format: (_, environment) => formatEnvironmentDates(environment.createdAt),
```
martinboulais, line 55: Same

lib/public/views/Environments/format/formatEnvironmentDates.js
```diff
@@ -0,0 +1,33 @@
...
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { h } from '/js/src/index.js';
+import { getLocaleDateAndTime } from '../../../utilities/dateUtils.js';
+
+/**
+ * Format a given run's start date
```
martinboulais, line 18: Please, avoid blind copy paste...

lib/public/views/Environments/format/formatEnvironmentDates.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param {number|null} timestamp the timestamp to format (in milliseconds)
+ * @return {Component} the formatted date
+ */
+export const formatEnvironmentDates = (timestamp) => {
```
martinboulais, line 23: Before to create a new function, try to find if an already existing one may do the job. In this case, if you have a look at formatTimestamp, you will see that you can have it display on 2 lines

lib/public/views/Environments/format/formatEnvironmentDates.js
```diff
@@ -0,0 +1,33 @@
...
+    const { date, time } = getLocaleDateAndTime(timestamp);
+    return h(
+        '',
+        h('', date),
+        h(
+            '.flex-row.g2.items-center',
+            [h('', time)],
```
martinboulais, line 30: No need to use flex if you have only one child. Same, no need to use an array. Again, please do not blindly copy-paste existing code

martinboulais (changes requested):

lib/public/app.css
mennokonijn: I tried using this, but due to the fact that all statuses are separate components instead of one text it is not possible to use this

martinboulais: Ok, that makes sense. In that case, please use `em` or `rem` at least.
Also, use more explicit class: `table-cl-mh-2` is completely opaque to me

martinboulais (changes requested):

martinboulais:
On medium screen, the id column is shrink. I would reduce the size of the current status to 10, the status history to 20

martinboulais (dismissed):

PR #1327, opened 2024-01-15
[O2B-1104] environment history occurrences API

SandroBoelsz:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- None

- The statistics API now contains an ' /environments/historyOccurrences' endpoint, which can be used to retrieve the most frequent environment history combinations and their occurrences within a given timeframe.

Changes made to the database:
- None


lib/database/repositories/EnvironmentHistoryItemRepository.js
```diff
@@ -11,8 +11,35 @@
...
+    SELECT environments, COUNT(*) AS count
+    FROM (
+        SELECT e.id, GROUP_CONCAT(ehi.status) AS environments
+        FROM environments e
+        JOIN environments_history_items ehi ON e.id = ehi.environment_id
+        JOIN (
+            SELECT DISTINCT environment_id
+            FROM environments_history_items
+            WHERE created_at BETWEEN '${timestampToMysql(from)}' AND '${timestampToMysql(to)}'
+        ) AS environment_ids_in_period
```
martinboulais, line 35: Actually, I think you should symply check that the environment has `created_at` before the `to` and `updated_at` after the `from`

lib/database/repositories/EnvironmentHistoryItemRepository.js
```diff
@@ -11,8 +11,35 @@
...
+ * Returns the SQL query for fetching the environment history combinations and
+ * their occurrences within a specified time period.
+ *
+ * @param {Period} period the period on which the occurrences must be counted
+ * @return {string} the raw SQL query
+ */
+const getHistoryDistributionQuery = ({ from, to }) => `
+    SELECT environments, COUNT(*) AS count
+    FROM (
+        SELECT e.id, GROUP_CONCAT(ehi.status) AS environments
```
martinboulais, line 28: I would rather name it `statusHistory` than `environment`

lib/database/repositories/EnvironmentHistoryItemRepository.js
```diff
@@ -34,6 +61,17 @@ class EnvironmentHistoryItemRepository extends Repository {
     async bulkInsert(historyItems) {
         return EnvironmentHistoryItem.bulkCreate(historyItems, { returning: false });
     }
+
+    /**
+     * Returns the occurrences of each environment history combination for a given period.
+     *
+     * @param {Period} period the period on which occurrences must be counted
+     * @return {Promise<{environments: string, count: number}[]>} the resulting histogram
```
martinboulais, line 69: It is not really environment, but statusHistory

martinboulais (changes requested):

lib/database/repositories/EnvironmentHistoryItemRepository.js
SandroBoelsz: You're entirely right. Now that I think about this approach, it should be more efficient and less complicated.

lib/database/repositories/EnvironmentHistoryItemRepository.js
SandroBoelsz: Again you are entirely right, the only reason why I put it like this, is because of the slides that refer to statusHistory as environments.

lib/database/repositories/EnvironmentHistoryItemRepository.js
SandroBoelsz: You're right, it was the same reason as the comment above.

martinboulais (approved):

PR #1328, opened 2024-01-15
[O2B-1103] Environment history occurrences histogram

SandroBoelsz:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- The statistics page now contains a histogram which displays the most frequent environment history combinations and their occurrences within a given timeframe.
-  Changed existing environment status acronyms in accordance to the RMreportW29_2023

- Changed the naming for the existing displayEnvironmentStatusHistory.js file to displayEnvironmentStatusHistoryOverview.js and used the displayEnvironmentStatusHistory.js filename for the display of the histogram environment status.

Changes made to the database:
- None


martinboulais:
image
Looks good in general. However: could you have a look to display all the labels (like for tags occurrences)? I would suggest to display a limited amount of bins, for example ~32 as a start, (at some point we may think about defining this parameter depending on the size of the screen but not for now), and maybe put some "other" category, or simply nothing (I'm affraid other would make a huge bar at the start...).

martinboulais (changes requested):
See https://github.com/AliceO2Group/Bookkeeping/pull/1328#issuecomment-1898089761

martinboulais (approved):

PR #1337, opened 2024-01-16
[O2B-1123] Use generic overview model in runs model

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- NA

- Refactor RunsOverviewModel to use OverviewPageModel

Changes made to the database:
- NA


lib/public/views/Runs/Overview/exportRunsTriggerAndModal.js
```diff
@@ -18,25 +18,29 @@ import { runsActiveColumns } from '../ActiveColumns/runsActiveColumns.js';
...
+    const loadingData = !runs;
+    const exportEnabled = selectedRunsFields.length > 0 && !loadingData;
+
+    // eslint-disable-next-line require-jsdoc
+    const optionalTruncatedInfo = () => runsOverviewModel.areExportableItemsTruncated
+        ? h(
+            '#truncated-export-warning.warning',
+            `The runs export is limited to ${runs.length} entries, only the last runs will be exported (sorted by run number)`,
+        )
+        : null;
```
xsalonx, line 40: This message is not render immediately, I think during `loadingData`, spinner might be here instead

lib/public/views/Runs/Overview/RunsOverviewModel.js
```diff
@@ -231,9 +211,9 @@ export class RunsOverviewModel extends Observable {
             || this.environmentIdsFilter !== ''
             || this.runQualitiesFilters.length !== 0
             || this._triggerValuesFilters.size !== 0
-            || this.nDetectorsFilter !== null
-            || this._nEpnsFilter !== null
-            || this.nFlpsFilter !== null
+            || this.nDetectorsFilter?.operator && this.nDetectorsFilter.limit
+            || this._nEpnsFilter?.operator && this._nEpnsFilter?.limit
+            || this.nFlpsFilter?.operator && this.nFlpsFilter?.limit
```
martinboulais, line 216: Do not belong to this PR

lib/public/views/Runs/Overview/exportRunsTriggerAndModal.js
martinboulais, line 40: Please, stop disable eslint. In particular when the function name is not really explicit.
I don't see what extracting this function brings here, especially if it's at the same level as where the code were called before.

lib/public/views/Runs/Overview/exportRunsTriggerAndModal.js
```diff
@@ -97,15 +101,15 @@ const errorDisplay = () => h('.danger', 'Data fetching failed');
...
+    const runsRemoteData = runsOverviewModel.exportableItems;
 
     return h('div#export-runs-modal', [
         h('h2', 'Export Runs'),
         runsRemoteData.match({
-            NotAsked: () => errorDisplay(),
             Loading: () => exportForm(runsOverviewModel, null, modalHandler),
             Success: (payload) => exportForm(runsOverviewModel, payload, modalHandler),
-            Failure: () => errorDisplay(),
+            Other: () => errorDisplay(),
```
martinboulais, line 112: Are you sure this is a good idea? I think `NotAsked` displaying an error was a mistake, and you are hiding it here. `Other` should not lead to some error display.

lib/public/views/Runs/Overview/RunsOverviewModel.js
```diff
@@ -51,81 +52,59 @@ export class RunsOverviewModel extends Observable {
...
 
-        this._pagination = new PaginationModel();
-        this._pagination.observe(() => this.fetchRuns());
-        this._pagination.itemsPerPageSelector$.observe(() => this.notify());
-
-        // Content
-        this._currentPageRuns = RemoteData.notAsked();
-        this._allRuns = RemoteData.notAsked();
+        // Exportable items
+        this._exportableDataSource = new PaginatedRemoteDataSource();
```
martinboulais, line 56: Why is it paginated?

lib/public/views/Runs/Overview/RunsOverviewModel.js
```diff
@@ -798,38 +778,42 @@ export class RunsOverviewModel extends Observable {
...
+     * Update the cache containing all the runs without paging
      *
-     * @return {RemoteData} the remote data of the runs
+     * @return {Promise<void>} void
+     * @private
      */
-    get allRuns() {
-        if (this._allRuns.isNotAsked()) {
-            this._fetchAllRunsWithoutPagination();
+    async loadExportableRuns() {
```
martinboulais, line 795: You are completely breaking the previous logic here, are you aware of it?
Before, we fetch all runs only if the current list of runs can't be reused as export.

lib/public/views/Runs/Overview/RunsOverviewModel.js
```diff
@@ -798,38 +778,42 @@ export class RunsOverviewModel extends Observable {
...
+        const endpoint = this.getRootEndpoint();
+        if (this._exportableItemsEndpoint === endpoint &&
+            (this.exportableItems.isSuccess() || this.exportableItems.isLoading())) {
```
martinboulais, line 798: You are using deprecated code in new code, you should not do that.

lib/public/views/Runs/Overview/RunsOverviewModel.js
```diff
@@ -51,81 +52,59 @@ export class RunsOverviewModel extends Observable {
...
-        this._pagination = new PaginationModel();
-        this._pagination.observe(() => this.fetchRuns());
-        this._pagination.itemsPerPageSelector$.observe(() => this.notify());
-
-        // Content
-        this._currentPageRuns = RemoteData.notAsked();
-        this._allRuns = RemoteData.notAsked();
+        // Exportable items
+        this._exportableDataSource = new PaginatedRemoteDataSource();
+        this._exportableItems = ObservableData.builder()
```
martinboulais, line 57: For me it should simply be `exportItems`, because it's the items of the export.

martinboulais (changes requested):

martinboulais:
I think you are doing too much things here: 
- Update filtering 
- Use OverviewModel
- Use data source for runs export

This PR should only contain code change for the 2nd point, please update accordingly

lib/public/views/Runs/Overview/RunsOverviewModel.js
xsalonx: revoked

lib/public/views/Runs/Overview/exportRunsTriggerAndModal.js
xsalonx: changes, revoked

lib/public/views/Runs/Overview/exportRunsTriggerAndModal.js
xsalonx: revoked, if using `NotAsked` was an error, changing it doesn't belong to this PR...

lib/public/views/Runs/Overview/RunsOverviewModel.js
xsalonx: bad idea, change to just `RemoteDataSource` with proper `RemoteData#apply`

lib/public/views/Runs/Overview/RunsOverviewModel.js
xsalonx: refactored to use match

lib/public/views/Runs/Overview/RunsOverviewModel.js
xsalonx: appied

lib/public/views/Runs/Overview/RunsOverviewModel.js
xsalonx: and it is so now, you can't reuse iff you changed filtering and this iff `root endpoint` is different - see condition if `loadExportRuns` (previously: `loadExportableRuns`)

martinboulais (approved):

PR #1342, opened 2024-01-17
[O2B-1108] Environment filter on current status

CollinHva:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- none

- environment API now allows filtering on current status

Changes made to the database:
- none


lib/usecases/environment/GetAllEnvironmentsUseCase.js
```diff
@@ -58,14 +58,26 @@ class GetAllEnvironmentsUseCase {
...
+        // Apply these filters after getting all environments to avoid unnecessary code for getting environment status history
+        if (filter?.currentStatus) {
+            const filters = filter.currentStatus.split(',').map((id) => id.trim());
+            rows = rows.filter((environment) => {
+                const { historyItems } = environment;
+                const lastHistoryItem = historyItems?.length > 0 ? historyItems[historyItems.length - 1] : null;
+                return filters.includes(lastHistoryItem?.status);
+            });
+            count = rows.length;
+        }
```
martinboulais, line 79: Because it is an association, you will indeed have to do the querry in two steps: first fetch the environment by filtering them, and then fetch the associations.
This way, you will probably not need `separate: true` btw

The filtering have to be done in sequelize query, because it will be much quicker

martinboulais (changes requested):

martinboulais (approved):

PR #1354, opened 2024-01-19
[O2B-1128] Implemented status history legend on environment overview

SandroBoelsz:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- When you hover over the status history on the environment overview page, a tooltip is now available, presenting a legend that provides explanations for all the status history acronyms.

- None

Changes made to the database:
- None


lib/public/views/Environments/format/displayEnvironmentStatusHistoryLegend.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+export const displayEnvironmentStatusHistoryLegend = (trigger) =>
+    tooltip(trigger, h('', [
+        h('h5', 'Status History Legend'),
+        Object.keys(STATUS_ACRONYMS).map((status) =>
+            h('.flex-row.justify-between', [
+                h('', status),
+                h('', STATUS_ACRONYMS[status]),
+            ])),
+    ]));
```
martinboulais, line 32: I think it is not correct to have the legend being a component that wraps the actual history (in trigger).
I would simply have the environmentStatusHistoryLegend to return the actual body of the tooltip, because maybe you will want to display this somewhere else, not in a popover.
Do not put `display` in function name, you can put Component as suffix if you like.

What you can do is have a component like `environmentStatusHistory` that display the history plus the popover, and this component call your function as the trigger body.

martinboulais (changes requested):

lib/public/views/Environments/format/displayEnvironmentStatusHistoryLegend.js
SandroBoelsz: Done, your suggestions would definitely improve the reusability.

lib/public/components/environments/environmentStatusHistoryComponent.js
```diff
@@ -0,0 +1,24 @@
...
+import { environmentStatusHistoryLegendComponent } from './environmentStatusHistoryLegendComponent.js';
+import { tooltip } from '../common/popover/tooltip.js';
+
+/**
+ * Returns a tooltip legend for the environment status history acronyms.
+ *
+ * @param {Component} trigger the element which will display the status history legend when hovered
+ * @return {Component} the resulting environment status history component
+ */
+export const environmentStatusHistoryComponent = (trigger) =>
```
martinboulais, line 23: Here no need to ask for trigger, you should simply expect the history items as parameter and completely hide the popover thing

martinboulais (changes requested):

lib/public/components/environments/environmentStatusHistoryComponent.js
```diff
@@ -0,0 +1,50 @@
...
+ * @return {Component} the resulting environment status history component
+ */
+export const environmentStatusHistoryComponent = (statusHistory) => tooltip(
+    h(
+        '.flex-row',
+        statusHistory.map((value, index) =>
+            h('.flex-row', [
+                coloredEnvironmentStatusComponent(value.status, value.acronym),
+                h('', index < statusHistory.length - 1 ? '-' : ''),
+            ])),
```
xsalonx, line 47: I think it can be quite simplified:
```
statusHistory.map(({ status, acronym }, index) => [
            coloredEnvironmentStatusComponent(status, acronym),
            index < statusHistory.length - 1 ? '-' : '',
        ]).flat(),
```
other option is to do it with a reduce method

lib/public/views/Environments/format/displayEnvironmentStatusHistoryOverview.js
```diff
@@ -25,13 +25,10 @@ export const displayEnvironmentStatusHistoryOverview = (environment) => {
...
-        .map(({ status }) => ({ status, content: STATUS_ACRONYMS[status] }));
+        .map(({ status }) => ({ status, acronym: STATUS_ACRONYMS[status] }));
 
     return h(
         '.flex-row',
-        statusHistory.map((value, index) => h('.flex-row', [
-            coloredEnvironmentStatusComponent(value.status, value.content),
-            h('', index < statusHistory.length - 1 ? '-' : ''),
-        ])),
+        environmentStatusHistoryComponent(statusHistory),
```
xsalonx, line 32: I think it might be good to put something like
```
h(
        '.flex-row',
        h(
            '.w-25',
            environmentStatusHistoryComponent(statusHistory),
        ),
    );
```
in order to widen hooverable area, then you can have cursor beyond area of the content

xsalonx: Besides, personally, tooltip on right hand side looks better - areas of the content and tooltip are not so close, but, as you prefer

lib/public/components/environments/environmentStatusHistoryComponent.js
martinboulais: I am pretty sure I had an example for this, but I can't find it...
The best is to use flatMap, I'll push the change

lib/public/views/Environments/format/displayEnvironmentStatusHistoryOverview.js
martinboulais: I really don't understand the reason of the `w-25` ? Why would you append 75% of blank space to the column?

lib/public/components/environments/environmentStatusHistoryComponent.js
martinboulais: Done, thanks for noticing

lib/public/views/Environments/format/displayEnvironmentStatusHistoryOverview.js
xsalonx: Well it might be w-15, w-50 whatever, I just wanted to suggest to widen area you can hoover in order to get tooltip, because with current solution, you need to hoover over status-items, which I consider not very user friendly, as well as having whole column "hooverable"

xsalonx: [](url)
Screenshot 2024-01-25 at 11 14 31
Screenshot 2024-01-25 at 11 13 00

xsalonx: hmm, I didn't capture my cursor, but it's supposed to be next to left border of tooltip at upper screenshot

martinboulais: Adding w-something will reduce the width of the cell's content to a percentage of the cell, which for me is a terrible idea because you force a white space at the end if the history item is long.
The point you raised is good though, I changed the way it's done by displaying an info icon at the top.

martinboulais (approved):

xsalonx (approved):

PR #1360, opened 2024-01-22
[O2B-988] Sort DPL tasks alphabetically

galaxywolfv:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
Ensure alphabetical sorting for DPL tasks.

### DIFF
--- a/lib/server/services/dpl/DplProcessService.js
+++ b/lib/server/services/dpl/DplProcessService.js
@@ -77,13 +77,7 @@ class DplProcessService {
             raw: true,
             attributes: ['id'],
         })).map(({ id }) => id);
-
-        const detectors = await DplDetectorRepository.findAll({ where: { id: { [Op.in]: ids } } });
-        const sortedDetectors = detectors
-            .map(dplDetectorAdapter.toEntity)
-            .sort((a, b) => a.name.localeCompare(b.name));
-
-        return sortedDetectors;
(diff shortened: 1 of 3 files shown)


lib/server/services/dpl/DplProcessService.js
```diff
@@ -77,7 +77,13 @@ class DplProcessService {
...
             attributes: ['id'],
         })).map(({ id }) => id);
-        return (await DplDetectorRepository.findAll({ where: { id: { [Op.in]: ids } } })).map(dplDetectorAdapter.toEntity);
+
+        const detectors = await DplDetectorRepository.findAll({ where: { id: { [Op.in]: ids } } });
+        const sortedDetectors = detectors
+            .map(dplDetectorAdapter.toEntity)
+            .sort((a, b) => a.name.localeCompare(b.name));
+
+        return sortedDetectors;
```
martinboulais, line 86: You are on the right tracks. However, this works in this particular case, but it doesn't if we apply pagination.
For example, if you have 3 items, let's say

```js
id: 1, name: 'aaa',
id: 2, name: 'ccc',
id: 3, name: 'bbb',
```
If you get the first two items then sort, you will have `aaa` and `ccc`, but you would expect `aaa`, `bbb`.

To solve this, you should apply ordering directly in the query, using https://sequelize.org/docs/v6/core-concepts/model-querying-basics/#ordering

This of course apply to the other modifications you did.

martinboulais (changes requested):

martinboulais (approved):

PR #1363, opened 2024-01-23
[O2B-907] Add color to tags

mennokonijn:
[02B-667]
[02B-907]
[02B-908]
[02B-909]

- [x] PR labels are selected

Notable changes for users:
- new feature to add a color to tags
- new feature to see color for tags

- 

Changes made to the database:
- extra column 'color in tags table'


lib/public/views/Tags/Create/index.js
```diff
@@ -62,6 +62,23 @@ const tagCreationForm = ({ data }) => {
...
+            maxlength: 6,
+            oninput: (e) => {
+                if (e.target.value !== '') {
+                    data.color = e.target.value;
+                } else {
+                    data.color = null;
+                }
+            },
+            style: { backgroundColor: `#${color}` },
+        }),
```
martinboulais, line 81: You can extract a `colorInputComponent` from this. BTW, you actually have a color HTML input (you'll have to remove the `#` from its value actually but it's better to use it

martinboulais (changes requested):

lib/public/views/Runs/ActiveColumns/runsActiveColumns.js
```diff
@@ -102,8 +103,8 @@ export const runsActiveColumns = {
         visible: true,
         profiles: [profiles.none, 'lhcFill', 'environment'],
         classes: 'w-5 f6',
-        format: (tags) => tags && tags.length > 0 ? tags.map(({ text }) => text).join(', ') : '-',
-        exportFormat: (tags) => tags && tags.length > 0 ? tags.map(({ text }) => text).join(', ') : '-',
+        format: (tags) => tags && tags.length > 0 ? formatTagListColor(tags) : '-',
+        exportFormat: (tags) => tags && tags.length > 0 ? formatTagListColor(tags) : '-',
```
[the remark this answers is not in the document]
martinboulais: The `exportFormat` should not have color

lib/public/components/common/form/inputs/colorInputComponent.js
```diff
@@ -0,0 +1,28 @@
...
+ * Returns a create button component
+ *
+ * @param {string} color The current color go
+ * @param {function} onInput The function of what must happen on input
+ * @return {Component} The color input element
+ */
+export const colorInputComponent = (color, onInput) => h('input#color.form-control.w-10.mb2', {
+    type: 'color',
+    value: color,
+    oninput: onInput,
```
martinboulais, line 26: The onInput callback should directly return the HEXA code, removing the `#`

lib/public/components/tag/tagDetail.js
```diff
@@ -115,21 +116,10 @@ const activeFields = (detailsModel) => ({
...
-                oninput: (e) => {
-                    if (e.target.value !== '') {
-                        detailsModel.setTagChange('color', e.target.value);
-                    } else {
-                        detailsModel.setTagChange('color', null);
-                    }
-                },
-                style: { backgroundColor: `#${detailsModel.tagChanges['color']}` },
+            colorInputComponent(detailsModel.tagChanges['color'], (e) => {
+                detailsModel.setTagChange('color', (e.target.value || null).replace(/^#/, ''));
```
martinboulais, line 120: (e.target.value || null).replace will throw if e.target.value is falsy

lib/public/components/tag/tagDetail.js
```diff
@@ -115,21 +116,10 @@ const activeFields = (detailsModel) => ({
...
             }) :
-            h('.rounded-1', { style: { backgroundColor: `#${color}` } }, color ? `#${color}` : ''),
+            h('.br2', { style: { backgroundColor: `#${color}` } }, color ? `#${color}` : ''),
```
martinboulais, line 122: If you have no color, backgroundColor will be '#' which is invalid, you should include the style internary too.

lib/server/controllers/tags.controller.js
```diff
@@ -43,6 +43,9 @@ const createTag = async (request, response) => {
     if (!value) {
         return;
     }
+    if (value.body.color?.length === 6) {
```
graduta, line 46: The `colorDto` which is part of `CreateTagDto` already checks for a length of 6.
Do we need to check again here?

lib/domain/dtos/common/colorDto.js
graduta, line 1: I see that all `DTO`s are named with upper cammel case, including those in common folder.  
Should this not be the case for this one as well?

lib/server/controllers/tags.controller.js
```diff
@@ -291,6 +294,11 @@ const updateTagById = async (request, response) => {
     if (!value) {
         return;
     }
+
+    if (value.body.color?.length === 6) {
```
graduta, line 298: Same comment as above as `UpdateTagDto` also uses `colorDto`

lib/public/components/common/form/inputs/colorInputComponent.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+export const colorInputComponent = (color, onInput) => h('.flex-row.items-center.g2', [
+    h('input#color', {
+        type: 'color',
+        value: color,
+        oninput: (e) => onInput(e.target.value ?? null),
+    }),
+    h(
+        '.btn.btn-pill.f7',
+        { disabled: !color, onclick: () => onInput(null) },
```
graduta, line 31: Perhaps a title or suggestion is also added to that `x` buttons refers to.
It is not very intuitive that it `resets` the selection. For me while I had the color palette opened, I thought it would close the color palette.

graduta (changes requested):
I believe it would be useful to also add the colored tags in the `log creation page`
While I do see the changes being applied in the `Views/Logs/Create`, the changes do not take effect

lib/server/controllers/tags.controller.js
martinboulais: Updated

lib/domain/dtos/common/colorDto.js
martinboulais: Yes you're right, updated

lib/server/controllers/tags.controller.js
martinboulais: Done

lib/public/components/common/form/inputs/colorInputComponent.js
martinboulais: I added a title

PR #1378, opened 2024-02-01
[O2B-1106] Preliminary cleaning of log creation component

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

- Slightly cleaned log creation component


graduta (changes requested):
When creating a log, and then trying to create another one, the `content` field does not get reset and contains the same values as before.

martinboulais:
> When creating a log, and then trying to create another one, the `content` field does not get reset and contains the same values as before.

Nice catch. This is not new to this PR, but it's fixed by the next PR in line

graduta (approved):

PR #1411, opened 2024-02-16
[O2B-904] Add API endpoint to fetch a list of main logs

AnoeBanoe:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- none

- New endpoint for fetching root logs only created using services

Changes made to the database:
- none


test/api/logs.test.js
```diff
@@ -774,6 +774,49 @@ module.exports = () => {
...
+                        return;
+                    }
+
+                    expect(res.body.data.every((log) => log.parentLogId === null));
+
+                    done();
+                });
+        });
+
+        it('should return an array', (done) => {
```
martinboulais, line 794: You should put more explicit name, such as "should successfully return paginated logs" then based on seeders, you check different pagination (and you check the log ids)

lib/server/controllers/logs.controller.js
```diff
@@ -281,6 +283,40 @@ const getLogById = async (request, response, next) => {
...
+
+            await response.status(200).json({
+                data: logs,
+                meta: {
+                    page: {
+                        pageCount: totalPages,
+                        totalCount: count,
+                    },
+                },
+            });
```
martinboulais, line 313: We have a function to not have to do that manually: `countedItemsToHttpView`
You can have a look and try to use it there.

lib/server/controllers/logs.controller.js
```diff
@@ -281,6 +283,40 @@ const getLogById = async (request, response, next) => {
...
 
+/**
+ * Endpoint to list all the root logs
+ *
+ * @param {Object} request the express request
+ * @param {Object} response the express response
+ * @return {void}
+ */
+const getRootLogs = async (request, response) => {
+    const value = await dtoValidator(GetAllLogsDto, request, response);
```
martinboulais, line 294: In your case, you don't have parameters such as filter that you may find in GetAllLogsDto. You should build a custom DTO to simply look for pagination.
You can find an example here: https://github.com/AliceO2Group/Bookkeeping/blob/9114239e18201fbb83fc9dfc513aef6144c3a87e/lib/server/controllers/dataPasses.controller.js
By removing the filtering option too.

lib/server/controllers/logs.controller.js
```diff
@@ -281,6 +283,40 @@ const getLogById = async (request, response, next) => {
...
+
+    if (value) {
+        try {
+            const { query: { page: { limit = ApiConfig.pagination.limit } = {} } } = value;
+
+            // Get the logs for the current page
+            const { count, logs } = await logService.getAllRootLogs({ ...value, query: { ...value.query, page: { limit } } });
```
martinboulais, line 301: the services should have no notion of query, it's business layer only (query is an API notion).
Here, what you need to pass to service is directly the offset and limit.

lib/server/controllers/logs.controller.js
```diff
@@ -281,6 +283,40 @@ const getLogById = async (request, response, next) => {
...
 };
 
+/**
+ * Endpoint to list all the root logs
+ *
+ * @param {Object} request the express request
+ * @param {Object} response the express response
+ * @return {void}
+ */
+const getRootLogs = async (request, response) => {
```
martinboulais, line 293: Controller functions are suffixed by `Handler`

lib/server/services/log/LogService.js
```diff
@@ -47,6 +49,36 @@ class LogService {
...
+            const { query = {} } = dto;
+            const { page = {} } = query;
+
+            queryBuilder.include('user')
+                .include('tags')
+                .include({ association: 'runs', attributes: ['id', 'runNumber'] })
+                .include({ association: 'lhcFills', attributes: ['fillNumber'] })
+                .include('subsystems')
+                .include('attachments')
+                .include({ association: 'environments', attributes: ['id'] });
```
martinboulais, line 68: If you extract log ids only, you probably don't need any include here.

lib/server/services/log/LogService.js
```diff
@@ -47,6 +49,36 @@ class LogService {
...
+
+            const { limit = ApiConfig.pagination.limit, offset = 0 } = page;
```
martinboulais, line 70: You should not use something called `ApiConfig` in the business layer, setting this default belongs to the controller.

lib/server/services/log/getRootLogIds.js
```diff
@@ -0,0 +1,44 @@
...
+
+    // Map over the logs to extract only the IDs
+    const logIds = logs.map((log) => log.id);
+
+    // Return the count and the extracted IDs
+    return {
+        count: result.count,
+        logIds: logIds,
+    };
+};
```
martinboulais, line 44: TBH here I would directly include this in the main service (if at some point you need to reuse this in another service, it may be worth extracting it).

lib/server/services/log/getRootLogIds.js
```diff
@@ -0,0 +1,44 @@
...
+    if (qbConfiguration) {
+        qbConfiguration(queryBuilder);
+    }
+
+    // Getting all logs first
+    const result = await LogRepository.findAndCountAll(queryBuilder);
+    const logs = result.rows;
+
+    // Map over the logs to extract only the IDs
+    const logIds = logs.map((log) => log.id);
```
martinboulais, line 37: If you only use one of the attribute I would suggest to specify it in the list of attributes you expect from sequelize and use raw querry, to avoid sequelize process of hydrating models: you can have a look at an example here https://github.com/AliceO2Group/Bookkeeping/blob/52a06143d4d64bb5d7b0e6562df980a389b0e8e7/lib/server/services/dpl/DplProcessService.js#L115

Also, you should use object destructuring here too:
```js
const logIds = logs.map(({id}) => id);
```

lib/server/services/log/getRootLogIds.js
```diff
@@ -0,0 +1,44 @@
...
+
+    queryBuilder.where('parentLogId').is(null);
+
+    if (qbConfiguration) {
+        qbConfiguration(queryBuilder);
+    }
+
+    // Getting all logs first
+    const result = await LogRepository.findAndCountAll(queryBuilder);
+    const logs = result.rows;
```
martinboulais, line 34: You can use object destructuring here:
```js
const {logs} = await LogRepository.findAndCountAll(queryBuilder);
```

martinboulais (changes requested):

lib/server/controllers/logs.controller.js
AnoeBanoe: done

lib/server/controllers/logs.controller.js
AnoeBanoe: done

lib/server/controllers/logs.controller.js
AnoeBanoe: done

lib/server/services/log/LogService.js
AnoeBanoe: done

lib/server/services/log/LogService.js
AnoeBanoe: done

lib/server/services/log/getRootLogIds.js
AnoeBanoe: done

lib/server/services/log/getRootLogIds.js
AnoeBanoe: done

test/api/logs.test.js
AnoeBanoe: done

lib/server/services/log/getRootLogIds.js
AnoeBanoe: done

lib/server/controllers/logs.controller.js
AnoeBanoe: done

lib/database/repositories/Repository.js
```diff
@@ -53,15 +53,11 @@ class Repository {
     /**
      * Returns all entities.
      *
-     * @param {QueryBuilder} [queryBuilder] The QueryBuilder to use.
+     * @param {QueryBuilder} [findQuery] The QueryBuilder to use.
```
martinboulais, line 56: When you update a parameter, take consideration of its documentation, it's almost as important as the code itself!

martinboulais (changes requested):

lib/database/repositories/Repository.js
AnoeBanoe: done

lib/server/controllers/logs.controller.js
```diff
@@ -281,6 +286,40 @@ const getLogById = async (request, response, next) => {
...
+ * Get all root logs
+ *
+ * @param {Object} request The *request* object represents the HTTP request and has properties for the request query
+ *                         string, parameters, body, HTTP headers, and so on.
+ * @param {Object} response The *response* object represents the HTTP response that an Express app sends when it gets an
+ *                          HTTP request.
+ * @returns {undefined}
+ */
+const getRootLogsHandler = async (request, response) => {
+    const value = await dtoValidator(
```
martinboulais, line 299: Your variable names should always mean things. For example here, I would suggest `requestDto` instead of value.

lib/server/services/log/LogService.js
```diff
@@ -47,6 +47,44 @@ class LogService {
...
+    }
+
+    /**
+     * Fetches the IDs of all root logs within specified pagination limits and the total count of all root logs.
+     * @param {number} limit The maximum number of log IDs to return in a single call.
+     * @param {number} offset The number of log IDs to skip before starting to return results.
+     * @return {Promise<{logIds: unknown[], count: number}>} A promise that resolves with an object containing an array of root log IDs
+     * for the current page and the total count of all root logs, or null if no logs are found.
+     */
+    async getRootLogIds(limit, offset) {
```
martinboulais, line 74: I think this function can be private to the service, so by convention its name should start by `_`.

```suggestion
    async _getRootLogIds(limit, offset) {
```

Being private, you don't need to test it

lib/server/services/log/LogService.js
```diff
@@ -47,6 +47,44 @@ class LogService {
...
 
+    /**
+     * Retrieves a paginated list of all root logs along with the total count.
+     * @param {number} limit The maximum number of root logs to retrieve in a single call.
+     * @param {number} offset The number of root logs to skip before starting to return results.
+     * @return {Promise<{count: number, logs: Log[]}>} A promise that resolves with an object containing the total count of all root logs,
+     * and an array of the root logs for the current page, or null if no logs are found.
+     */
+    async getAllRootLogs(limit, offset) {
+        const { count, logIds } = await this.getRootLogIds(limit, offset);
```
martinboulais, line 58:
```suggestion
        const { count, logIds } = await this._getRootLogIds(limit, offset);
```

lib/server/services/log/LogService.js
```diff
@@ -47,6 +47,44 @@ class LogService {
...
     }
 
+    /**
+     * Retrieves a paginated list of all root logs along with the total count.
+     * @param {number} limit The maximum number of root logs to retrieve in a single call.
+     * @param {number} offset The number of root logs to skip before starting to return results.
+     * @return {Promise<{count: number, logs: Log[]}>} A promise that resolves with an object containing the total count of all root logs,
+     * and an array of the root logs for the current page, or null if no logs are found.
+     */
+    async getAllRootLogs(limit, offset) {
```
martinboulais, line 57: Please add a few tests for this function

martinboulais (changes requested):

lib/server/controllers/logs.controller.js
AnoeBanoe: done

lib/server/services/log/LogService.js
AnoeBanoe: done

lib/server/services/log/LogService.js
AnoeBanoe: done

test/lib/server/services/log/LogService.test.js
```diff
@@ -20,13 +20,16 @@ const { expect } = chai;
...
+    let rootLogs;
 
     beforeEach(async () => {
         logMock = {
             title: 'Yet another log',
             text: 'This is the text field of yet another log.',
             tags: [],
         };
+
+        rootLogs = await logService.getAllRootLogs(10, 0);
```
martinboulais, line 32: in before each, you set up things that should be done for every tests, for example setting mock data.
Getting rootLogs is actually part of your tests, so you should write it in your tests when it applies.

test/lib/server/services/log/LogService.test.js
```diff
@@ -47,4 +50,12 @@ module.exports = () => {
...
         expect(logsForFill.map(({ id }) => id)).to.eql([1, 3, 4]);
     });
+
+    it('Should successfully return the amount of root logs based off the limit', async () => {
+        expect(rootLogs.logs).to.lengthOf(10);
+    });
+
+    it('Should successfully return only root logs', async () => {
+        rootLogs.logs.forEach((log) => expect(log.parentLogId).to.equal(log.id));
+    });
```
martinboulais, line 60: These can be merged in one. What you can do however, is that the pagination and offset works fine.

martinboulais (changes requested):

test/lib/server/services/log/LogService.test.js
AnoeBanoe: done

martinboulais (approved):

PR #1412, opened 2024-02-16
[O2B-1161] Store only PHYSICS runs from MonALISA

xsalonx:
#### I have a JIRA ticket
- [x] branch and/or PR name(s) include(s) JIRA ID
- [x] issue has "Fix version" assigned
- [x] issue "Status" is set to "In review"
- [x] PR labels are selected

Notable changes for users:
- NA

Notable changes for developers:
- Add constraint on inserting data from MonALISA - only PHYSICS runs association are stored now

Changes made to the database:
- NA

PR #1413, opened 2024-02-16
[O2B-1162] Display all PHYSICS runs on Data Pass view

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- All PHYSICS runs are displayed on Runs Per Data Pass view, regardless of their online quality

- Use breadcrumbs for Runs Per Data Pass page title and tooltip to display error

Changes made to the database:
- NA


xsalonx:
To be merged after https://github.com/AliceO2Group/Bookkeeping/pull/1412

lib/public/components/common/navigation/breadcrumbs.js
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+import { h, iconChevronRight } from '/js/src/index.js';
+
+/**
+ * Create Breadcrumbs out of provided items, each items is preceded by rightChevron Icon
+ * @param {Component[]} items items in breadcrumbs
+ * @return {Component} breadcrumbs
+ */
+export const breadcrumbs = (...items) => h('.flex-row.g1.items-center', items.flatMap((item) => [iconChevronRight(), item]));
```
martinboulais, line 22: Why `...` items and not simply take an array?
I think that you should not put chevron in the first component, the breadcrumbs should include all the items

lib/public/views/Runs/RunPerDataPass/RunsPerDataPassOverviewPage.js
```diff
@@ -47,13 +50,14 @@ export const RunsPerDataPassOverviewPage = ({ runs: { perDataPassOverviewModel }
...
-                NotAsked: () => h('h2', 'Good, physics runs'),
-            }),
+        h('.flex-row.justify-between.items-center.g2', [
+            h('h2', 'Physics Runs'),
+            h('.flex-row', dataPass.match({
+                Success: (payload) => breadcrumbs(h('h2', payload.name)),
+                Failure: (payload) => tooltip(iconWarning(), errorAlert(payload)),
+                Loading: () => spinner({ size: 2, absolute: false }),
+                NotAsked: () => tooltip(iconWarning(), errorAlert([{ detail: 'No data was asked for' }])),
+            })),
```
martinboulais, line 60: I think that breadcrumbs should include the `Physics runs`. I would rather create a variable `commonTitle = h('h2', 'Physics Runs')` and use it in every cases

Also, if you test the code you'll see that the warning is very small, I would put it in `.f3` for example

martinboulais (changes requested):

lib/public/views/Runs/RunPerDataPass/RunsPerDataPassOverviewPage.js
xsalonx: applied

lib/public/components/common/navigation/breadcrumbs.js
xsalonx: 1. refactored breadcrumbs so it doesn't put chevron as first element.
2. I would keep parameters with spread operator, as there is no particular reason to use array, it would required one additional code element - array constructor, it would be more handy if breadcrumbs were supposed to work with many elements, naturally being held in array, but it's not

lib/public/views/Runs/RunPerDataPass/RunsPerDataPassOverviewPage.js
```diff
@@ -46,14 +49,19 @@ export const RunsPerDataPassOverviewPage = ({ runs: { perDataPassOverviewModel }
...
-                NotAsked: () => h('h2', 'Good, physics runs'),
-            }),
+        h('.flex-row.justify-between.items-center.g2', [
+            h(
+                '.flex-row.g1.items-center',
+                dataPass.match({
+                    Success: (payload) => breadcrumbs(commonTitle, h('h2', payload.name)),
+                    Failure: (payload) => [commonTitle, tooltip(h('.f3', iconWarning()), errorAlert(payload))],
+                    Loading: () => [commonTitle, spinner({ size: 2, absolute: false })],
+                    NotAsked: () => [commonTitle, tooltip(h('.f3', iconWarning()), errorAlert([{ detail: 'No data was asked for' }]))],
```
martinboulais, line 62: 2 things:
- I would not use errorAlert in tooltip, sorry if I was unclear about it, I think that a simple text is enough
- The message brings no info at all, it was better before, like `Failed to fetch data pass name` for example...

martinboulais (changes requested):

lib/public/components/common/navigation/breadcrumbs.js
xsalonx: refactored not to use spread operator

lib/public/views/Runs/RunPerDataPass/RunsPerDataPassOverviewPage.js
xsalonx: For not asked change to simple text, for Failure I'd keep `errorAlert` as it's only handy method to display actual error

lib/public/components/common/navigation/breadcrumbs.js
```diff
@@ -15,8 +15,8 @@
 import { h, iconChevronRight } from '/js/src/index.js';
 
 /**
- * Create Breadcrumbs out of provided items, each items is preceded by rightChevron Icon
+ * Create Breadcrumbs out of provided items which are separated with rightChevron Icon
  * @param {Component[]} items items in breadcrumbs
  * @return {Component} breadcrumbs
  */
-export const breadcrumbs = (...items) => h('.flex-row.g1.items-center', items.flatMap((item) => [iconChevronRight(), item]));
+export const breadcrumbs = (items) => h('.flex-row.g1.items-center', items.flatMap((item) => [iconChevronRight(), item]).slice(1));
```
martinboulais, line 22: This solution is indeed more elegant than mine. Could you please also update the other places where we do the other way aound? In this PR is fine, there is not much and this has no impact.

martinboulais (changes requested):

lib/public/components/common/navigation/breadcrumbs.js
xsalonx: applied

martinboulais (approved):

PR #1423, opened 2024-02-26
[O2B-905] Add endpoint to fetch all replies to a given log using its id

AnoeBanoe:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- none

- New endpoint added that uses a log id to fetch it's child logs

### DIFF
--- a/lib/server/controllers/logs.controller.js
+++ b/lib/server/controllers/logs.controller.js
@@ -37,6 +37,7 @@ const { updateExpressResponseFromNativeError } = require('../express/updateExpre
 const { countedItemsToHttpView } = require('../utilities/countedItemsToHttpView');
 const { PaginationDto } = require('../../domain/dtos');
 const { DtoFactory } = require('../../domain/dtos/DtoFactory');
+const EntityIdDto = require('../../domain/dtos/EntityIdDto');
 
 /**
  * Create a new log
@@ -320,12 +321,18 @@ const getRootLogsHandler = async (request, response) => {
  * Get all child logs by parent id
  */
 const getChildLogsByParentIdHandler = async (request, response) => {
-    const requestDto = await dtoValidator(GetLogDto, request, response);
+    const requestDto = await dtoValidator(
+        DtoFactory.paramsOnly({
+            logId: EntityIdDto,
+        }),
+        request,
+        response,
+    );
 
     if (requestDto) {
         try {
-            const { params: { logId } } = requestDto;
-            const childLogs = await logService.getChildLogsByParentId(logId);
+            const { params: { logId: parentLogId } } = requestDto;
+            const childLogs = await logService.getChildLogsByParentId(parentLogId);
 
             await response.status(200).json({
                 data: childLogs,
--- a/lib/server/controllers/logs.controller.js
+++ b/lib/server/controllers/logs.controller.js
@@ -286,14 +286,9 @@ const getLogById = async (request, response, next) => {
     }
 };
 
+// eslint-disable-next-line valid-jsdoc
 /**
  * Get all root logs
- *
- * @param {Object} request The *request* object represents the HTTP request and has properties for the request query
- *                         string, parameters, body, HTTP headers, and so on.
- * @param {Object} response The *response* object represents the HTTP response that an Express app sends when it gets an
- *                          HTTP request.
- * @returns {undefined}
  */
 const getRootLogsHandler = async (request, response) => {
     const requestDto = await dtoValidator(
@@ -320,14 +315,9 @@ const getRootLogsHandler = async (request, response) => {
     }
 };
 
+// eslint-disable-next-line valid-jsdoc
 /**
  * Get all child logs by parent id
- *
- * @param {Object} request The *request* object represents the HTTP request and has properties for the request query
- *                         string, parameters, body, HTTP headers, and so on.
- * @param {Object} response The *response* object represents the HTTP response that an Express app sends when it gets an
- *                          HTTP request.
- * @returns {undefined}
  */
 const getChildLogsByParentIdHandler = async (request, response) => {
     const requestDto = await dtoValidator(GetLogDto, request, response);
@@ -337,21 +327,10 @@ const getChildLogsByParentIdHandler = async (request, response) => {
             const { params: { logId } } = requestDto;
             const { count, childLogs } = await logService.getChildLogsByParentId(logId);
 
-            if (count > 0) {
-                await response.status(200).json({
-                    data: childLogs,
-                    totalCount: count,
-                });
-            } else {
-                await response.status(404).json({
-                    errors: [
-                        {
-                            status: '404',
-                            title: `Child logs of log (${request.params.logId}) could not be found`,
-                        },
-                    ],
-                });
-            }
+            await response.status(200).json({
+                data: childLogs,
+                totalCount: count,
+            });
         } catch (error) {
             updateExpressResponseFromNativeError(response, error);
         }
--- a/lib/server/controllers/logs.controller.js
+++ b/lib/server/controllers/logs.controller.js
@@ -325,11 +325,10 @@ const getChildLogsByParentIdHandler = async (request, response) => {
     if (requestDto) {
         try {
             const { params: { logId } } = requestDto;
-            const { count, childLogs } = await logService.getChildLogsByParentId(logId);
+            const childLogs = await logService.getChildLogsByParentId(logId);
 
             await response.status(200).json({
                 data: childLogs,
-                totalCount: count,
             });
         } catch (error) {
             updateExpressResponseFromNativeError(response, error);
--- a/lib/server/controllers/logs.controller.js
+++ b/lib/server/controllers/logs.controller.js
@@ -320,6 +320,44 @@ const getRootLogsHandler = async (request, response) => {
     }
 };
 
+/**
+ * Get all child logs by parent id
+ *
+ * @param {Object} request The *request* object represents the HTTP request and has properties for the request query
+ *                         string, parameters, body, HTTP headers, and so on.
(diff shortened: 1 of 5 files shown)


lib/server/controllers/logs.controller.js
```diff
@@ -320,6 +320,44 @@ const getRootLogsHandler = async (request, response) => {
...
 };
 
+/**
+ * Get all child logs by parent id
+ *
+ * @param {Object} request The *request* object represents the HTTP request and has properties for the request query
+ *                         string, parameters, body, HTTP headers, and so on.
+ * @param {Object} response The *response* object represents the HTTP response that an Express app sends when it gets an
+ *                          HTTP request.
+ * @returns {undefined}
```
martinboulais, line 330: We don't define all this in controller anymore, we simply put what the route does (what you did well line 324) and we mark it as ignored by eslint. Please have a look in `dplProcess.controller.js`

lib/server/controllers/logs.controller.js
```diff
@@ -320,6 +320,44 @@ const getRootLogsHandler = async (request, response) => {
...
+ */
+const getChildLogsByParentIdHandler = async (request, response) => {
+    const requestDto = await dtoValidator(GetLogDto, request, response);
+
+    if (requestDto) {
+        try {
+            const { params: { logId } } = requestDto;
+            const { count, childLogs } = await logService.getChildLogsByParentId(logId);
+
+            if (count > 0) {
```
martinboulais, line 340: Why would you throw if count is 0? In this case, you should simply return an empty, which is normal.

lib/server/services/log/LogService.js
```diff
@@ -85,6 +85,40 @@ class LogService {
...
     }
 
+    /**
+     * Retrieves a list of replies to a log based off the given logId
+     * @param {number} logId The id of the log of which the child logs are retrieved
+     * @return {Promise<{count: number, childLogs: Log[]}>} A promise that resolves with an object containing the total count of all child logs,
+     * and an array of the child logs for the given parent log
+     */
+    async getChildLogsByParentId(logId) {
+        const { count, childLogIds } = await this._getChildLogIds(logId);
```
martinboulais, line 95: I am not sure why you fetch id first here? I think that you can directly fetch the logs no?

test/api/logs.test.js
```diff
@@ -1551,4 +1554,56 @@ module.exports = () => {
...
+            request(server)
+                .get(`/api/logs/${logWithChildrenId}/children`)
+                .expect(200)
+                .end((err, res) => {
+                    if (err) {
+                        done(err);
+                        return;
+                    }
+
+                    expect(res.body.data.every((log) => log.parentLogId === logWithChildrenId));
```
martinboulais, line 1569: You should also check the amount of items here

martinboulais (changes requested):

lib/server/controllers/logs.controller.js
AnoeBanoe: Done

lib/server/services/log/LogService.js
AnoeBanoe: I do that so I can use _getAllByIds, which has some extra functionalities. Would it be better to just implement those functionalities directly and fetch the logs instead of id's?

test/lib/server/services/log/LogService.test.js
```diff
@@ -66,4 +66,32 @@ module.exports = () => {
         expect(rootLogsFirstPage.logs[0].id).to.equal(1);
         expect(rootLogsSecondPage.logs[0].id).to.equal(5);
     });
+
+    it('Should successfully return child logs of their parent log', async () => {
+        const logWithChildrenId = 117;
+        const { childLogs } = await logService.getChildLogsByParentId(logWithChildrenId);
+
+        childLogs.forEach((log) => expect(log.parentLogId === logWithChildrenId));
```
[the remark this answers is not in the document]
AnoeBanoe: done

test/lib/server/services/log/LogService.test.js
```diff
@@ -66,4 +66,32 @@ module.exports = () => {
...
+
+        expect(childLogs).to.be.empty;
+    });
+
+    it('Should successfully return correct count when children are found', async () => {
+        const logWithChildrenId = 117;
+        const { count, childLogs } = await logService.getChildLogsByParentId(logWithChildrenId);
+
+        expect(childLogs.length).to.equal(count);
+    });
```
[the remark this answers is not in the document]
AnoeBanoe: I get count from findAndCountAll, so I thought this was simply the amount of logs since no limit or offset are given for these child logs. Am I mistaken?

test/lib/server/services/log/LogService.test.js
```diff
@@ -66,4 +66,32 @@ module.exports = () => {
...
+
+    it('Should successfully return correct count when children are not found', async () => {
+        const logWithoutChildrenId = 122;
+        const { count, childLogs } = await logService.getChildLogsByParentId(logWithoutChildrenId);
+
+        expect(childLogs.length).to.equal(count);
```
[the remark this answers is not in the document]
AnoeBanoe: done

test/api/logs.test.js
AnoeBanoe: done

lib/server/controllers/logs.controller.js
martinboulais: Yes I would say so

test/lib/server/services/log/LogService.test.js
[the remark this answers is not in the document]
martinboulais: Ah ok I did not notice. Then I think you should not return the count and simply use `findAll` in here, and you should put a hard-coded length in the test.

lib/server/services/log/LogService.js
```diff
@@ -85,6 +85,20 @@ class LogService {
         return { count, logIds };
     }
 
+    /**
+     * Retrieves a list of replies to a log based off the given logId
+     * @param {number} logId The id of the log of which the child logs are retrieved
+     * @return {Promise<{childLogs: Log[]}>} A promise that resolves with an object containing
+     * an array of the child logs for the given parent log
```
martinboulais, line 92: Don't forget to update the type of return, it's not a object with `childLogs` anymore 🙂 

Also, I think that you can simplify a bit the text, something like `Resolves with the replies logs` would be clear enough I think.

lib/server/services/log/LogService.js
```diff
@@ -131,17 +145,36 @@ class LogService {
...
      */
     async _getAllByIds(logIds) {
-        const logQueryBuilder = dataSource.createQueryBuilder()
-            .where('id').oneOf(...logIds)
-            .include({ association: 'runs', attributes: ['id', 'runNumber'] })
-            .include({ association: 'lhcFills', attributes: ['fillNumber'] })
-            .include({ association: 'environments', attributes: ['id'] });
+        const logQueryBuilder = this._createFindQueryBuilder();
 
+        logQueryBuilder.where('id').oneOf(...logIds);
```
martinboulais, line 150: I forgot to mention, but here you can chain the where: 
```js
const logQueryBuilder = this._createFindQueryBuilder()
    .where('id').oneOf(...logIds);
```
If you think this way is less clear, keep the one you have.

lib/server/services/log/LogService.js
```diff
@@ -131,17 +145,36 @@ class LogService {
...
         const logs = await LogRepository.findAll(logQueryBuilder);
+        return this._prepareLogsEntities(logs);
+    }
+
+    /**
+     * Return all the logs after preparing them to entities and adding childrenCount
+     *
+     * @param {logs[]} logs the list of logs to fetch
```
martinboulais, line 158: logs to fetch? Rather to prepare no?
The type is `SequelizeLog[]` not `logs[]`

lib/server/controllers/logs.controller.js
```diff
@@ -320,6 +315,27 @@ const getRootLogsHandler = async (request, response) => {
     }
 };
 
+// eslint-disable-next-line valid-jsdoc
+/**
+ * Get all child logs by parent id
+ */
+const getChildLogsByParentIdHandler = async (request, response) => {
+    const requestDto = await dtoValidator(GetLogDto, request, response);
```
martinboulais, line 323: I apologize, I did not notice at first, but here you should not use the same DTO, because its purpose is not to get one log by id. 
Here you can directly generate the DTO on the fly using DtoFactory as you did well for root logs.
I would also rather name the parameter `parentLogId` instead of `logId`

martinboulais (changes requested):

lib/server/controllers/logs.controller.js
```diff
@@ -320,6 +315,28 @@ const getRootLogsHandler = async (request, response) => {
...
+
+    if (requestDto) {
+        try {
+            const { params: { logId } } = requestDto;
+            const { count, childLogs } = await logService.getChildLogsByParentId(logId);
+
+            await response.status(200).json({
+                data: childLogs,
+                totalCount: count,
```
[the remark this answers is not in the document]
AnoeBanoe: This should be closed already I think?

lib/server/services/log/LogService.js
AnoeBanoe: done

lib/server/services/log/LogService.js
AnoeBanoe: I think the chaining is clear, done.

lib/server/services/log/LogService.js
AnoeBanoe: done

lib/server/controllers/logs.controller.js
AnoeBanoe: done

PR #1425, opened 2024-02-26
[O2B-1137] Make RCT Runs overviews horizontally scrollable

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Runs Per Data Pass, Runs Per LHC Period and Runs per Simulation Pass views are horizontally scrollable

- Extended capabilities of table component:
    - It can be configured as horizontally scrollable

Changes made to the database:
- NA


lib/public/app.css
```diff
@@ -673,3 +673,27 @@ label {
...
+    background-color: var(--color-gray-light);
+}
+
+.x-scrollable .table.left-sticky tbody tr td:first-child {
+    position: sticky;
+    top: 0;
+    left: 0;
+    z-index: 2;
+}
+
```
martinboulais, line 696: This is a hack, and I don't know if it is a good idea to propose it as is, because it will stay and they will soon expect more. We should really try to propose only stable and clean stuff.
For example, what will you do when they will want 2 columns to be sticky?
Also, we should not have something like `left-sticky` to set z-index, and even less a background color... this will lead to maintenance hell!
For me, a first step is to put horizontal scroll, and then we see when reworking table to put the sticky column. But we should not provide such hack.

martinboulais (changes requested):
Please, again, stop rushing! On my screen, this PR breaks table display by removing the shadow...

lib/public/app.css
xsalonx: 1. At first I did it with modification of `rows`, `cell` components then with a bit of refactoring it could be more elastic and be used for more than one frozen (sticky) column, there is need only to one sticky column for, I'd say, primary-key column, which is always leftmost one.
2. I think it's quite natural to set z-index with sticky in this case, because leftmost column must be always on top of everything on right, what's
Screenshot 2024-02-26 at 16 27 07
 not the case without setting z-index. 
3. I dared to bring background colour here because:
    - it's to preserve common (used everywhere) styling of table
    - without this leftmost column is a bit hollow (see screenshot below)
    - for me, without sticky column, that scrolling is not very useful because you don't know run it is while looking one some quality
image

image

lib/public/models/OverviewModel.js
```diff
@@ -168,6 +170,23 @@ export class OverviewPageModel extends Observable {
         return this._pagination;
     }
 
+    /**
+     * Get value stating whether table should be horizontally scrollable
+     * @return {boolean} whether display is scrollable
+     */
+    get displayXScrollable() {
```
martinboulais, line 177: This is my personal feeling, I would rather name it "isHorizontalScrollEnabled" and "setHorizontalScrollEnabled"

lib/public/components/common/table/table.js
```diff
@@ -123,7 +124,7 @@ export const table = (
     }
 
     // Extract the profile of the table
-    const { profile: currentProfile = profiles.none } = tableConfiguration || {};
+    const { profile: currentProfile = profiles.none, xScrollable = false } = tableConfiguration || {};
```
martinboulais, line 127: This is my personal feeling, I would rather name it "horizontalScroll"

lib/public/views/Runs/RunPerDataPass/RunsPerDataPassOverviewPage.js
```diff
@@ -81,10 +86,22 @@ export const RunsPerDataPassOverviewPage = ({ runs: { perDataPassOverviewModel }
...
+
             exportRunsTriggerAndModal(perDataPassOverviewModel, modalModel),
         ]),
         h('.flex-column.w-100', [
-            table(remoteRuns, activeColumns, null, { profile: 'runsPerDataPass' }),
+            table(
+                remoteDetectors.match({
+                    Success: () => remoteRuns,
+                    Other: () => remoteDetectors,
+                }),
```
martinboulais, line 97: Why is that? Besides, it do not belongs here

lib/public/views/Runs/RunPerPeriod/RunsPerLhcPeriodOverviewPage.js
```diff
@@ -60,12 +65,23 @@ export const RunsPerLhcPeriodOverviewPage = ({ runs: { perLhcPeriodOverviewModel
...
             h('h2', `Good, physics runs of ${lhcPeriodName}`),
             exportRunsTriggerAndModal(perLhcPeriodOverviewModel, modalModel),
         ]),
         h('.flex-column.w-100', [
-            table(remoteDetectors.isSuccess() ? remoteRuns : remoteDetectors, activeColumns, null, { profile: 'runsPerLhcPeriod' }),
+            table(
+                remoteDetectors.match({
+                    Success: () => remoteRuns,
+                    Other: () => remoteDetectors,
+                }),
```
martinboulais, line 77: Same comment

lib/public/views/Runs/RunsPerSimulationPass/RunsPerSimulationPassOverviewPage.js
```diff
@@ -84,7 +89,18 @@ export const RunsPerSimulationPassOverviewPage = ({
             exportRunsTriggerAndModal(runsPerSimulationPassOverviewModel, modalModel),
         ]),
         h('.flex-column.w-100', [
-            table(remoteRuns, activeColumns, null, { profile: 'runsPerSimulationPass' }),
+            table(
+                remoteDetectors.match({
+                    Success: () => remoteRuns,
+                    Other: () => remoteDetectors,
+                }),
```
martinboulais, line 96: Same here

martinboulais (dismissed):

lib/public/components/common/table/table.js
```diff
@@ -101,6 +101,7 @@ const parseColumnsConfiguration = (columns, currentProfile) => {
  * @param {Object|null} [tableConfiguration] the global table configuration
  * @param {string} [tableConfiguration.profile] the profile to apply to the table, to filter out columns or apply
  *     specific configuration. If not specified, any visible column will be displayed
+ * @param {boolean} [tableConfiguration.xScrollable = false] is true use x scroll in case of overflow, fixed layout otherwise
```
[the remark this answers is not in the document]
xsalonx: fixed

lib/public/models/OverviewModel.js
```diff
@@ -168,6 +170,23 @@ export class OverviewPageModel extends Observable {
...
+        return this._displayXScrollable;
+    }
+
+    /**
+     * Set value stating whether table should be horizontally scrollable
+     * @param {Boolean} displayXScrollable whether display is scrollabe
```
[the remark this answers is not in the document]
xsalonx: fixed

lib/public/models/OverviewModel.js
```diff
@@ -168,6 +170,23 @@ export class OverviewPageModel extends Observable {
...
+     */
+    set displayXScrollable(displayXScrollable) {
+        this._displayXScrollable = Boolean(displayXScrollable);
```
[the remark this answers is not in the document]
xsalonx: applied

lib/public/models/OverviewModel.js
xsalonx: applied

lib/public/components/common/table/table.js
xsalonx: applied

lib/public/views/Runs/RunPerDataPass/RunsPerDataPassOverviewPage.js
xsalonx: because previous was wrong, anyway it's revoked

lib/public/views/Runs/RunsPerSimulationPass/RunsPerSimulationPassOverviewPage.js
xsalonx: because previous was wrong, anyway it's revoked

lib/public/views/Runs/RunPerPeriod/RunsPerLhcPeriodOverviewPage.js
```diff
@@ -60,12 +65,20 @@ export const RunsPerLhcPeriodOverviewPage = ({ runs: { perLhcPeriodOverviewModel
     };
 
     return h('', [
-        h('.flex-row.justify-between.items-center', [
+        h('.flex-row.justify-between.items-center.g2', [
```
graduta, line 68: Is there a need for adding `g2`? I would argue due to the usage of `justify-between`, you do not need either `items-center` or `g2`

lib/public/components/common/table/table.js
```diff
@@ -135,10 +137,10 @@ export const table = (
         Error(`Unhandled type <${typeof data}> of data : ${data ? JSON.stringify(data) : data}`);
     }
 
-    return h(
+    return h(`${horizontalScrollEnabled ? '.scroll-auto' : ''}`, h(
```
graduta, line 140: With the addition of a new `h` element, you are losing the UI delimiter of the table: `shadow-level` which makes it difficult from a visual point to see where the table ends. 

Your changes:
image

Current behaviour:
image

lib/public/views/Runs/Overview/RunsOverviewModel.js
```diff
@@ -59,6 +59,8 @@ export class RunsOverviewModel extends OverviewPageModel {
         // Export items
         this._allRuns = RemoteData.NotAsked();
 
+        this._horizontalScrollEnabled = false;
```
graduta, line 62: What is the purpose of this line if the variable is defined by default to `false` in the OverviewModel?

graduta (changes requested):
I do not think this is the correct way to go as in this proposal, one has to modify both the model and the call to `table()` component in order to get the horizontal.

I believe a better way would be for the model to include the table configuration fields that is passed to the table. In this way, with one change (in the model) you can modify the table. 

e.g. right now in the run-overview table, one has to defined both in `RunOverviewModel` and `RunOverviewPage` this setting to get it working

lib/public/components/common/table/table.js
xsalonx: right, fixed

lib/public/views/Runs/RunPerPeriod/RunsPerLhcPeriodOverviewPage.js
xsalonx: `g2` removed, without `items-center` 'Export runs' goes into bad position

graduta (approved):

PR #1434, opened 2024-02-29
[0B-1152] auto fill environments and lhcFills in log reply

AnoeBanoe:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- When replying to a log, the environment and lhcFill id's are auto filled from their parent

- New default method expectInputValue, does the same as expectInnerText but then for values instead of innerHTML

### DIFF
--- a/test/public/defaults.js
+++ b/test/public/defaults.js
@@ -415,6 +415,28 @@ module.exports.fillInput = async (page, inputSelector, value, events = ['input']
 };
 
 /**
+ * Evaluate and return the value content of a given element handler
+ * @param {{evaluate}} elementHandler the puppeteer handler of the element to inspect
+ * @returns {Promise<XPathResult>} the html content
+ */
+const getInnerValue = async (elementHandler) => await elementHandler.evaluate((element) => element.value);
+
+module.exports.getInnerValue = getInnerValue;
+
+/**
+ * Expect an element to have a given value
+ *
+ * @param {Object} page Puppeteer page object.
+ * @param {string} selector Css selector.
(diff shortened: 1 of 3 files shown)


test/public/defaults.js
```diff
@@ -414,6 +414,28 @@ module.exports.fillInput = async (page, inputSelector, value, events = ['input']
     }, inputSelector, value, events);
 };
 
+/**
+ * Evaluate and return the value content of a given element handler
+ * @param {{evaluate}} elementHandler the puppeteer handler of the element to inspect
+ * @returns {Promise<XPathResult>} the html content
+ */
+const getInnerValue = async (elementHandler) => await elementHandler.evaluate((element) => element.value);
```
martinboulais, line 422: Very good idea! I would simply name it `getInputValue` and limit it to input

test/public/defaults.js
```diff
@@ -414,6 +414,28 @@ module.exports.fillInput = async (page, inputSelector, value, events = ['input']
...
+
+module.exports.getInnerValue = getInnerValue;
```
martinboulais, line 424: I think it's not useful to export it for now, you only use it internally

martinboulais (changes requested):

test/public/defaults.js
AnoeBanoe: done

test/public/defaults.js
AnoeBanoe: done

martinboulais (approved):

PR #1370, opened 2024-01-29
[O2B-1125] Implement Simulation Passes per Data Pass view

xsalonx:
#### I have a JIRA ticket
- [x] branch and/or PR name(s) include(s) JIRA ID
- [x] issue has "Fix version" assigned
- [x] issue "Status" is set to "In review"
- [x] PR labels are selected

Notable changes for users:
- Added Simulation Passes per Data Pass view

Notable changes for developers:
- NA

Changes made to the database:
- NA

PR #1436, opened 2024-03-01
[O2B-1178] Add counts of anchored simulation passes

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Added anchored MC counts to Data Pass per LHC Period view

- Refactored DataPassService - removed subqueries

Changes made to the database:
- NA


xsalonx:
To be merged after https://github.com/AliceO2Group/Bookkeeping/pull/1370

test/public/defaults.js
```diff
@@ -308,6 +309,19 @@ module.exports.expectInnerTextTo = async (page, selector, validator) => {
...
 };
 
+/**
+ * Expect an element to have a text valid against given validator
+ * @param {Object} page Puppeteer page object.
+ * @param {string} selector Css selector.
+ * @param {function<string, boolean>} validator text validator. It must return true if text is valid, retrun false or throw otherwise
+ * @return {Promise<void>} resolves once the text has been checked
+ */
+module.exports.expectInnerTextTo = async (page, selector, validator) => {
```
martinboulais, line 319: Wasn't this merged yet?

test/public/defaults.js
```diff
@@ -292,7 +292,8 @@ module.exports.getInnerText = getInnerText;
  */
 module.exports.expectInnerText = async (page, selector, innerText) => {
     await page.waitForSelector(selector, { timeout: 200 });
-    expect(await getInnerText(await page.$(selector))).to.equal(innerText);
+    const actualInnerText = await getInnerText(await page.$(selector));
+    expect(actualInnerText).to.equal(innerText);
```
martinboulais, line 296: No?

martinboulais (changes requested):

martinboulais:
Maybe add some test to check the real content of the column, meaning:
- check that the counter has the right value
- check that clicking on the link bring to the correct page

test/public/defaults.js
xsalonx: yes, some problem after merging main, checkout to main

xsalonx:
> Maybe add some test to check the real content of the column, meaning:
> 
> * check that the counter has the right value
> * check that clicking on the link bring to the correct page

Added

test/public/dataPasses/overviewPerLhcPeriod.test.js
```diff
@@ -67,6 +69,41 @@ module.exports = () => {
...
 
         await validateTableData(page, new Map(Object.entries(tableDataValidators)));
+        const names = await getAllDataFields(page, 'name');
+        const runs = await getAllDataFields(page, 'associatedRuns');
+        const anchoreds = await getAllDataFields(page, 'anchoredSimulationPasses');
+        expect(names.map((name, index) => ({
+            name,
+            runsCount: Number(runs[index].split('\n')[0]) || 0,
+            simulationPassesCount: Number(anchoreds[index].split('\n')[0]) || 0,
+        }))).to.have.all.deep.members([
```
martinboulais, line 79: This is a very far-fetched solution. It works, but getting data column by column, then build rows, and finally check it is hard to understand.
Can't you do `$eval` on rows, then inside it build an object returning the `{name, runsCount, simulationPassesCount}` using innerText?

test/public/dataPasses/overviewPerLhcPeriod.test.js
```diff
@@ -67,6 +69,41 @@ module.exports = () => {
...
+
+    it('can navigate to runs per data pass page', async () => {
+        await goToPage(page, 'data-passes-per-lhc-period-overview', { queryParameters: { lhcPeriodId: 2 } });
+        await waitForNavigation(page, () => pressElement(page, 'tbody tr td:nth-of-type(2)'));
+        const url = new URL(page.url());
+        expect(url.searchParams.get('page')).to.be.equal('runs-per-data-pass');
+        expect(Number(url.searchParams.get('dataPassId'))).to.be.a('Number');
+    });
+
+    it('can navigate to acnhored simulation passes per data pass page', async () => {
```
martinboulais, line 101: typo

test/public/dataPasses/overviewPerLhcPeriod.test.js
```diff
@@ -67,6 +69,41 @@ module.exports = () => {
...
+            },
+        ]);
+    });
+
+    it('can navigate to runs per data pass page', async () => {
+        await goToPage(page, 'data-passes-per-lhc-period-overview', { queryParameters: { lhcPeriodId: 2 } });
+        await waitForNavigation(page, () => pressElement(page, 'tbody tr td:nth-of-type(2)'));
+        const url = new URL(page.url());
+        expect(url.searchParams.get('page')).to.be.equal('runs-per-data-pass');
+        expect(Number(url.searchParams.get('dataPassId'))).to.be.a('Number');
```
martinboulais, line 98: Why not using `checkMismatchingUrlParam` ?

martinboulais (changes requested):

test/public/dataPasses/overviewPerLhcPeriod.test.js
xsalonx: fixed

test/public/dataPasses/overviewPerLhcPeriod.test.js
xsalonx: Because: checking correctness of the feature should concluded with checking current url, `mismatching stuff` would be helpful in case of checking exact match, for that I would need to have the `id`, which can me access only from the `model` (bad due to mixing layers). Other option is to fetch directly from endpoint, a bit cumbersome.

test/public/dataPasses/overviewPerLhcPeriod.test.js
xsalonx: refactored

test/public/defaults.js
```diff
@@ -280,13 +280,13 @@ module.exports.getAllDataFields = async (page, key) => {
...
-    for (const key of columnKeys) {
-        columnsData[key] = await this.getAllDataFields(page, key);
-    }
     const result = [];
-    for (let index = 0; index < columnsData[columnKeys[0]].length; index++) {
-        result.push(Object.fromEntries(columnKeys.map((key) => [key, columnsData[key][index]])));
+    for (const row of await page.$$('table tbody tr')) {
+        const entity = {};
+        for (const columnKey of columnKeys) {
+            entity[columnKey] = await row.$eval(`td[id$='${columnKey}']`, ({ innerText }) => innerText);
```
martinboulais, line 287: I would rather use `.column-${columnKey}`, in the end I think we should get rid of putting unique ids on every cells, what do you thinK?

martinboulais (changes requested):

test/public/dataPasses/overviewPerLhcPeriod.test.js
xsalonx: refactored to expect exact match

martinboulais (approved):

PR #1456, opened 2024-03-11
Reduced MonALISA Synchroniser error logs

xsalonx:
#### I DON'T have JIRA ticket
- [ ] tests are added

Notable changes for users:
- NA

- Added merging similar errors for MonALISA Synchroniser to prevent logs overflow

Changes made to the database:
- NA

### DIFF
--- a/lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
+++ b/lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
@@ -56,16 +56,21 @@ class MonAlisaSynchronizer {
 
             const errors = {};
             for (const monalisaDataPass of dataPasses) {
-                await dataSource.transaction(() => this._updateOneDataPass(monalisaDataPass))
-                    .catch((error) => {
-                        if (!errors[error.message]) {
-                            errors[error.message] = [];
-                        }
-                        errors[error.message].push(monalisaDataPass.name);
-                    });
+                try {
+                    await dataSource.transaction(() => this._updateOneDataPass(monalisaDataPass));
+                } catch (error) {
+                    if (!errors[error.message]) {
+                        errors[error.message] = [];
+                    }
+                    errors[error.message].push(monalisaDataPass.name);
+                }
             }
             if (Object.entries(errors).length !== 0) {
-                this.logger.errorMessage(JSON.stringify(errors), {});
+                const stackedError = Object.entries(errors)
+                    .map(([errorMessage, dataPassesList]) => `Error for data passes: ${dataPassesList.join(',')}\n   ${errorMessage}`)
+                    .join('\n');
+
+                this.logger.errorMessage(stackedError, {});
             }
             this.logger.debug('Synchronization of Data Passes completed');
         } catch (error) {
--- a/lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
+++ b/lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
@@ -54,26 +54,9 @@ class MonAlisaSynchronizer {
             const dataPasses = (await this._monAlisaInterface.getDataPasses())
                 .filter((monalisaDataPass) => this._doesDataPassNeedUpdate(monalisaDataPass, dataPassesLastRunNumber));
 
-            // eslint-disable-next-line require-jsdoc
-            const transactionablePipeline = async (monalisaDataPass) => {
-                const lhcPeriodDB = await getOrCreateLhcPeriod({ name: extractLhcPeriod(monalisaDataPass.name).name });
-                await DataPassRepository.upsert({ lhcPeriodId: lhcPeriodDB.id, ...monalisaDataPass });
-                const dataPass = await DataPassRepository.findOne({ where: { name: monalisaDataPass.name } });
-                const { runNumbers } = await this._monAlisaInterface.getDataPassDetails(dataPass.description);
-                const associatedRuns = await RunRepository.findAll({
-                    where: {
-                        runNumber: { [Op.in]: runNumbers },
-                        definition: RunDefinition.Physics,
-                    },
-                });
-                if (associatedRuns.length > 0) {
-                    await dataPass.addRuns(associatedRuns, { ignoreDuplicates: true });
-                }
-            };
-
             const errors = {};
             for (const monalisaDataPass of dataPasses) {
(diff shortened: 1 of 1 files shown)


lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
```diff
@@ -71,13 +71,22 @@ class MonAlisaSynchronizer {
...
+                    .catch((error) => {
+                        if (!errors[error.message]) {
+                            errors[error.message] = [];
+                        }
+                        errors[error.message].push(monalisaDataPass.name);
+                    });
+            }
+            if (Object.entries(errors).length !== 0) {
+                this.logger.errorMessage(JSON.stringify(errors), {});
             }
```
martinboulais, line 86: I think this is a strange way to proceed. If you have an error message very long, like an HTML error page, the error message you will display will be unreadable.
I would rather do it the following way:
Loop over the error messages, and put one log like: `Error occurred for data passes ${dataPasses.join(',')}:\n${message}`
Doing so, I would rather use a map than an object...

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
```diff
@@ -71,13 +71,22 @@ class MonAlisaSynchronizer {
                 }
             };
 
+            const errors = {};
             for (const monalisaDataPass of dataPasses) {
                 await dataSource.transaction(() => transactionablePipeline(monalisaDataPass))
```
martinboulais, line 76: Could you rename this `transactionablePipeline` by something that inform the user about what it does?
Please think thoughtfully the name you'll use.

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
```diff
@@ -71,13 +71,22 @@ class MonAlisaSynchronizer {
...
-                    .catch((error) => this.logger.errorMessage(error.message, {}));
+                    .catch((error) => {
```
martinboulais, line 77: You are mixing await with `.catch`, it's very strange. Please use only await.

martinboulais (commented):
Please try to split the big try in smaller tries to log specific error messages

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
```diff
@@ -71,13 +71,22 @@ class MonAlisaSynchronizer {
...
+                    });
+            }
+            if (Object.entries(errors).length !== 0) {
+                this.logger.errorMessage(JSON.stringify(errors), {});
             }
             this.logger.debug('Synchronization of Data Passes completed');
         } catch (error) {
-            this.logger.errorMessage(error.message, {});
+            this.logger.errorMessage(`Fatal error preventing synchronization of DataPasses: ${error.message}`, {});
             this.logger.debug('Synchronization of Data Passes failed');
```
martinboulais, line 90: Why is there this debug if you already logged an error message?

martinboulais (changes requested):

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
xsalonx: removed

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
xsalonx: applied

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
xsalonx: applied

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
xsalonx: refactored, used your formatting. and the longest expected error there is `Unexpected token '<', "<!doctype "... is not valid JSON`

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
```diff
@@ -54,31 +54,49 @@ class MonAlisaSynchronizer {
...
-                    await dataPass.addRuns(associatedRuns, { ignoreDuplicates: true });
+            const errors = {};
+            for (const monalisaDataPass of dataPasses) {
+                try {
+                    await dataSource.transaction(() => this._updateOneDataPass(monalisaDataPass));
+                } catch (error) {
+                    if (!errors[error.message]) {
+                        errors[error.message] = [];
+                    }
+                    errors[error.message].push(monalisaDataPass.name);
```
martinboulais, line 65:
```suggestion
                    const dataPassesNames = errors.get(error.message) ?? [];
                    if (dataPassesNames.length === 0) {
                        errors.set(error.message, dataPassesNames)
                    }
                    dataPassesNames.push(monalisaDataPass.name);
```

martinboulais: Using a map you can do this. I would also advise to rename the variable, errors is wrong, it is `dataPassesNamesMappedByErrorMessage` (or something else, but it's not errors that you are storing)

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
```diff
@@ -54,31 +54,49 @@ class MonAlisaSynchronizer {
...
                 }
-            };
+            }
+            if (Object.entries(errors).length !== 0) {
+                const stackedError = Object.entries(errors)
+                    .map(([errorMessage, dataPassesList]) => `Error for data passes: ${dataPassesList.join(',')}\n   ${errorMessage}`)
```
martinboulais, line 70: dataPassesList => dataPassesNames once again, naming is very important

martinboulais (changes requested):

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
xsalonx: renamed

lib/server/monalisa-synchronization/MonAlisaSynchronizer.js
xsalonx: removed

martinboulais (approved):

PR #1462, opened 2024-03-14
[O2B-1194] Refactor export functionality

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Changed export functionality:
    - Allow to input file name
    - Use spinner to show data are being loaded

- Extracted export-related methods to `OverviewPageModel`
- Generalised `exportTriggerAndModal` to work with another overview views

Changes made to the database:
- NA

### DIFF
--- a/lib/public/models/OverviewModel.js
+++ b/lib/public/models/OverviewModel.js
@@ -19,6 +19,11 @@ import { buildUrl } from '../utilities/fetch/buildUrl.js';
 import pick from '../utilities/pick.js';
 import { createCSVExport, createJSONExport } from '../utilities/export.js';
 
+export const EXPORT_FORMATS = {
+    JSON: 'JSON',
+    CSV: 'CSV',
+};
+
 /**
  * Interface of a model representing an overview page state
  *
@@ -62,8 +67,16 @@ export class OverviewPageModel extends Observable {
         this._exportObservableItems.bubbleTo(this);
         this._exportDataSource.pipe(this._exportObservableItems);
 
+        /**
+         * Default name for export data file, can be overriden in inherting classes
+         */
         this._defaultExportName = 'bkp-export';
         this._exportName = '';
+
+        /**
+         * Default format
+         */
+        this._defaultExportFormat = EXPORT_FORMATS.JSON;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


lib/public/models/OverviewModel.js
```diff
@@ -51,6 +58,25 @@ export class OverviewPageModel extends Observable {
 
         this._observableItems = ObservableData.builder().initialValue(RemoteData.loading()).build();
         this._observableItems.bubbleTo(this);
+
+        this._exportDataSource = new PaginatedRemoteDataSource();
```
martinboulais, line 62: Exports are not paginated

lib/public/models/OverviewModel.js
```diff
@@ -119,8 +145,97 @@ export class OverviewPageModel extends Observable {
...
+    /**
+     * Set export name
+     * @param {string} exportName name
+     */
+    set exportName(exportName) {
+        this._exportName = exportName;
+    }
+
+    /**
+     * Get the field values that will be exported
```
martinboulais, line 207: I think this comment is misleading, because it seems like you return the value of the fields.
I would rather use `Get the list of properties that will be exported for each items`

lib/public/models/OverviewModel.js
```diff
@@ -119,8 +145,97 @@ export class OverviewPageModel extends Observable {
...
+     * @return {string[]} the list of fields of a export items to be included in the export
+     */
+    get exportFields() {
+        return this._exportFields || [];
+    }
+
+    /**
+     * Set the selected fields to be exported
```
martinboulais, line 215: `Set the list of properties that will be exported for each items`

lib/public/models/OverviewModel.js
```diff
@@ -136,8 +251,30 @@ export class OverviewPageModel extends Observable {
     }
 
     /**
-     * Return the current items remote data
+     * States if the list of NOT paginated runs contains the full list of runs available under the given criteria
```
martinboulais, line 254: runs?

lib/public/models/OverviewModel.js
```diff
@@ -136,8 +251,30 @@ export class OverviewPageModel extends Observable {
...
      *
+     * @return {boolean|null} true if the runs list is not truncated (null if all items are not yet available)
+     */
+    get areExportItemsTruncated() {
+        return this.exportItems.match({
+            Success: (payload) => this.items.match({
```
martinboulais, line 260: Why do you need this second match?

lib/public/views/Runs/Overview/exportTriggerAndModal.js
```diff
@@ -0,0 +1,143 @@
+/**
```
martinboulais, line 1: I would go further and move this away from `Runs`

martinboulais (changes requested):
I would suggest a bit of rework on top of my comments:
- I propose to extract something like `ExportConfigurationModel` to store the configuration of the export, like selected columns and export type
- I would create two classes (and maybe a common interface, but not sure it makes sense without ts): something like `CsvExportGenerator` and `JsonExportGenerator` that both expose an `export` function that takes as parameter a list of items, a list of properties, a file name and generate the corresponding file.

PR #1489, opened 2024-04-03
[O2B-821] Make all public routes private 

graduta:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- HTTP API routes will become private, thus needing a token to be accessible

-

Changes made to the database:
-

### DIFF
--- a/lib/server/routers/runs.router.js
+++ b/lib/server/routers/runs.router.js
@@ -17,7 +17,7 @@ module.exports = {
     children: [
         {
             path: '/runs',
-            args: { public: true },
+            args: { public: false },
             children: [
                 {
                     method: 'get',
@@ -66,7 +66,7 @@ module.exports = {
             ],
         }, {
             path: '/legacy/runs',
-            args: { public: true },
+            args: { public: false },
             children: [
                 {
                     path: ':runId',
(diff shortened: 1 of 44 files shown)


lib/server/routers/runs.router.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
     children: [
         {
             path: '/runs',
-            args: { public: true },
```
martinboulais, line 20: Probably best is to simply put `args: {public: true}` on top and remove the sub-ones no?

martinboulais (changes requested):

lib/server/routers/runs.router.js
graduta: You mean `public: false` I guess?
This is the default behaviour of the server in any way, but I let it as false to be explicit

martinboulais: Yes public false.
TBH I would rather delete it from everywhere if it's the default, to me it looks like dead code.
But my comment was: removing it from children and putting it only at the top

graduta: I was not sure if you want to keep the convention as it was from the others.
Done, I removed all and let the default `visible` to be applied by the server.

martinboulais (dismissed):

martinboulais (approved):

PR #1508, opened 2024-04-12
[O2B-1186] QC flag type creation page improvements

xsalonx:
#### I have a JIRA ticket
- [ ] PR labels are selected

Notable changes for users:
- NA

- Fixed QC flag creation component scaling

Changes made to the database:
- NA


lib/public/app.css
```diff
@@ -81,6 +81,10 @@ html, body {
     width: unset;
 }
 
+.mw-large {
+    max-width: var(--ui-component-large);
+}
+
```
martinboulais, line 87: If you look at line 52, you'll see that:

- fixed width are put before percentage width, I would then put this line at l. 74
- we use `l` not `large`

lib/public/views/QcFlagTypes/Create/QcFlagTypeCreationPage.js
```diff
@@ -26,7 +26,7 @@ import { LabelPanelHeaderComponent } from '../../../components/common/panel/Labe
 const qcFlagTypeCreationComponent = (qcFlagTypeCreationModel) => {
     const { name, method, color, bad } = qcFlagTypeCreationModel.formData;
 
-    const panel = [
+    return h('.flex-column.gr3.mw-large', [
```
martinboulais, line 29: Why gr3? You can directly use g3

martinboulais (changes requested):

lib/public/app.css
xsalonx: applied

lib/public/views/QcFlagTypes/Create/QcFlagTypeCreationPage.js
xsalonx: applied

martinboulais (approved):

PR #1554, opened 2024-05-13
[O2B-1255] Disable a QC flag delete button when the flag is verified

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Disabled QC flag delete button for verified flags

- NA

Changes made to the database:
- NA

### DIFF
--- a/lib/public/views/QcFlags/details/qcFlagDetailsComponent.js
+++ b/lib/public/views/QcFlags/details/qcFlagDetailsComponent.js
@@ -105,7 +105,12 @@ export const qcFlagDetailsComponenet = (qcFlagDetailsModel, additionalDetailsCon
                     Loading: () => deleteButton(null, 'Processing...'),
                     Success: () => deleteButton(null, 'Processed!'),
                     Failure: () => deleteButton(() => qcFlagDetailsModel.delete()),
-                    NotAsked: () => deleteButton(() => qcFlagDetailsModel.delete()),
+                    NotAsked: () => remoteQcFlag.match({
+                        Success: (qcFlag) => qcFlag.verifications?.length > 0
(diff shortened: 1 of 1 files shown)


lib/public/views/QcFlags/details/qcFlagDetailsComponent.js
```diff
@@ -105,7 +105,12 @@ export const qcFlagDetailsComponenet = (qcFlagDetailsModel, additionalDetailsCon
                     Loading: () => deleteButton(null, 'Processing...'),
                     Success: () => deleteButton(null, 'Processed!'),
                     Failure: () => deleteButton(() => qcFlagDetailsModel.delete()),
-                    NotAsked: () => deleteButton(() => qcFlagDetailsModel.delete()),
+                    NotAsked: () => remoteQcFlag.match({
+                        Success: (qcFlag) => qcFlag.verifications?.length > 0
+                            ? tooltip(deleteButton(), 'Cannot delete verified flag')
```
graduta, line 110: Message is not descriptive enough. I would recommend something along the line:
`Deleting a verified flag is not allowed`

graduta (changes requested):

lib/public/views/QcFlags/details/qcFlagDetailsComponent.js
xsalonx: applied

graduta (approved):

graduta (changes requested):
I believe the `Failure` option should also check for verifications. Example of scenario:
- user 1 opens the qc flag details page
- user 2 opens the qc flag details page and verifies the flag
- user 1 wants to delete the flag but it fails due to back-end check. The failure message should check for verifications appeared in the mean time

PR #1559, opened 2024-05-17
[O2B-1241] RCT Runs overviews filtering

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Added filtering to Run Per Data Pass

- NA

Changes made to the database:
- NA


lib/public/components/Filters/common/filtersPanelPopover.js
```diff
@@ -14,6 +14,22 @@ import { popover } from '../../common/popover/popover.js';
...
+import { profiles } from '../../common/table/profiles.js';
+import { applyProfile } from '../../../utilities/applyProfile.js';
+
+/**
+ * Adapt column profiles for applyProfile function
+ *
+ * @param {string|string[]|object<string, *>} profilesToBeAdapted profiles
+ * @return {string[]|object<string, *>} adapted profile
+ */
+const adaptColumnProfiles = (profilesToBeAdapted) => {
```
martinboulais, line 26: Please, make an effort on naming and conventions!!!!
Adapt to what? What does it looks like, to what do you adapt it?
If a function make sense in a given context only (`applyProfile` function) then it must be inside of it, you must not tell in the documentation where the function must be used, it goes completely against the reusability common sense

lib/public/components/Filters/common/filtersPanelPopover.js
```diff
@@ -14,6 +14,22 @@ import { popover } from '../../common/popover/popover.js';
 import { PopoverTriggerPreConfiguration } from '../../common/popover/popoverPreConfigurations.js';
 import { PopoverAnchors } from '../../common/popover/PopoverEngine.js';
 import { h } from '/js/src/index.js';
+import { profiles } from '../../common/table/profiles.js';
+import { applyProfile } from '../../../utilities/applyProfile.js';
+
+/**
+ * Adapt column profiles for applyProfile function
+ *
+ * @param {string|string[]|object<string, *>} profilesToBeAdapted profiles
```
martinboulais, line 23: This is not documentation

lib/public/components/Filters/common/filtersPanelPopover.js
```diff
@@ -29,10 +45,15 @@ const filtersToggleTrigger = () => h('button#openFilterToggle.btn.btn.btn-primar
  *
  * @param {object} filteringModel the filtering model
  * @param {object} columns the list of columns containing filters
- *
+ * @param {object} [configuration] configuration
```
martinboulais, line 48: Missing description

lib/public/components/Filters/common/filtersPanelPopover.js
```diff
@@ -46,7 +67,9 @@ const filtersToggleContent = (filteringModel, columns) => h('.w-l.scroll-y.flex-
     ]),
     h('.flex-column.g2', [
         Object.entries(columns)
-            .filter(([_, column]) => column.filter)
+            .filter(([_, column]) =>column.filter
+                && (applyProfile(column, appliedProfile, adaptColumnProfiles(column.profiles))
+                 || (column.filterProfiles ?? []).includes(appliedProfile)))
```
martinboulais, line 72: Except if it's very clear, lambda should be on one line or have a body.
Here it would be way more clear to have:

However, I disagree with what you try to achieve here. The way profiles currently work is, if it's specified, profiles might disable columns or override their configuration. `filter` being one of the configuration properties, should be overridden in the profile property.
I agree that profiles system is flawed, and the whole active column system should be fixed, but adding this kind of dupplicated functionality can only make it worse.

test/public/defaults.js
```diff
@@ -510,11 +510,19 @@ module.exports.checkMismatchingUrlParam = async (page, expectedUrlParameters) =>
...
+            // Browser context, be careful when modifying
+            const names = [...document.querySelectorAll(`table tbody .column-${columnId}`)].map(({ innerText }) => innerText);
+            return JSON.stringify(names) === JSON.stringify(expectedInnerTextValues);
+        }, { timeout }, columnId, expectedInnerTextValues);
+    } catch {
+        throw new Error(`Column (${columnId}) is has incorrect values after timeout (${timeout}):
+            expected: [${expectedInnerTextValues}],
+            actual:   [${await this.getColumnCellsInnerTexts(page, columnId)}]
+        `);
+    }
```
martinboulais, line 525: The idea is very good, but I would rather not use `getColumnCellsInnerTexts`, you should get the value, compare it, then throw if they do not match using the value found in the first place. If you do it in two steps, the value can change in the mean time and provide a misleading error message

martinboulais (changes requested):

lib/public/components/Filters/common/filtersPanelPopover.js
xsalonx: removed

lib/public/components/Filters/common/filtersPanelPopover.js
xsalonx: removed

test/public/defaults.js
xsalonx: `getColumnCellsInnerTexts` cannot be called inside `waitForFunciton` and also cannot pass it's result to it.
Handle case when column has expected values only after timeout

lib/public/components/Filters/common/filtersPanelPopover.js
xsalonx: refactored

martinboulais (changes requested):

martinboulais (approved):

PR #1598, opened 2024-06-04
[O2B-1077] Fix again: keep scroll in popover

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- Clicking in popover do not reset the scroll anymore


lib/public/components/common/popover/PopoverEngine.js
```diff
@@ -151,7 +154,7 @@ class PopoverEngine {
         displayZoneMargins,
         position,
         projectors,
-        setChildrenSize,
+        configuration,
     ) {
```
xsalonx, line 158: If `configuration` is optional, it should be handled if undefined/null is provided

xsalonx (changes requested):

lib/public/components/common/popover/PopoverEngine.js
martinboulais: Here it's just transfered and the constructor accept undefined, so it's fine here

xsalonx: constructor doesn't accept undefined because it will throw `Uncaught TypeError: Cannot read properties of undefined (reading 'resetSize')`

martinboulais: Indeed, nice catch!

xsalonx (approved):

PR #1600, opened 2024-06-06
[O2B-1285] Differentiate the cases between Detector not involved and not yet given the flags

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Added iconBan to RCT runs overviews in case there is no QC flag assigned to detector and user has no permission to add one

- NA

Changes made to the database:
- NA


lib/public/views/Runs/ActiveColumns/runDetectorsAsyncQcActiveColumns.js
```diff
@@ -29,7 +30,8 @@ import { QC_SUMMARY_COLORS } from '../../../components/qcFlags/qcSummaryColors.j
  */
 const runDetectorAsyncQualityDisplay = (runNumber, dplDetectorId, { dataPassId, simulationPassId }, qcSummary) => {
     if (!qcSummary?.[runNumber]?.[dplDetectorId]) { // If there is no QC flag assigned
-        return qcFlagCreationPanelLink({ dataPassId, simulationPassId, runNumber, dplDetectorId });
+        return qcFlagCreationPanelLink({ dataPassId, simulationPassId, runNumber, dplDetectorId })
```
martinboulais, line 33: First, I think that such thing should go to to the `qcFlagCreationPanelLink`
Then, here you return the tooltip even in the case where dpl detector list is loading or has an error... (You don't use other in this case, but this can give the intuition why remote data cases should be handled carefully)

martinboulais (changes requested):

lib/public/views/Runs/ActiveColumns/runDetectorsAsyncQcActiveColumns.js
xsalonx: refactored

martinboulais (changes requested):

lib/public/views/Runs/ActiveColumns/runDetectorsAsyncQcActiveColumns.js
```diff
@@ -29,7 +30,15 @@ import { QC_SUMMARY_COLORS } from '../../../components/qcFlags/qcSummaryColors.j
...
 const runDetectorAsyncQualityDisplay = (runNumber, dplDetectorId, { dataPassId, simulationPassId }, qcSummary) => {
     if (!qcSummary?.[runNumber]?.[dplDetectorId]) { // If there is no QC flag assigned
-        return qcFlagCreationPanelLink({ dataPassId, simulationPassId, runNumber, dplDetectorId });
+        return qcFlagCreationPanelLink(
+            { dataPassId, simulationPassId, runNumber, dplDetectorId },
+            {
+                noPermissionContent: tooltip(
+                    h('.gray', iconBan()),
+                    'No QC flag was assigned. You have no permission to manage QC flag for this detector',
+                ),
```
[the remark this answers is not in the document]
xsalonx: The reason not to put it directly in function is that according previous recommendation nothing should be rendered in case user has no permissions on `QcFlagsFor(...)OverviewPage`. In case `noPermissionContent` is not passed as argument there it will be needed to duplicate logic of `remoteDplDetectors.match(...)` from `createRunDetectorsAsyncQcActiveColumns`. Besides, I added observableData with detectors which for user has expert permissions

lib/public/views/Runs/RunPerDataPass/RunsPerDataPassOverviewPage.js
```diff
@@ -65,20 +69,26 @@ export const RunsPerDataPassOverviewPage = ({ runs: { perDataPassOverviewModel }
...
-        ...createRunDetectorsAsyncQcActiveColumns(remoteDetectors.match({
-            Success: (payload) => payload,
-            Other: () => [],
-        }), {
-            profiles: 'runsPerDataPass',
-            dataPassId: remoteDataPass.match({
-                Success: (dataPass) => dataPass.id,
-                Other: () => null,
+        ...createRunDetectorsAsyncQcActiveColumns(
+            remoteDetectors.match({
```
martinboulais, line 73: Once again, I did not noticed the other in view.
This is once again bad, because if something goes wrong or is loading, the user is not informed about it.
If any of the remote data is failure, the page should probably display an error message. If it's loading, there should be a spinner somewhere (maybe a full page spinner, maybe only in the table).
This applies to all the remote data in this page.
Once again, you should almost NEVER use Other in views!

martinboulais (changes requested):

lib/public/Model.js
```diff
@@ -67,6 +71,17 @@ export default class Model extends Observable {
         // Bind user preferences
         userPreferencesStore.bubbleTo(this);
 
+        this._userDplDetectors = ObservableData.builder()
+            .source(dplDetectorsProvider.physical$)
+            .apply((remoteDetectors) => remoteDetectors.apply({
+                Success: (detectors) => {
+                    const { access = [] } = this.session;
+                    return detectors.filter(({ name }) => access.includes(BkpRoles.ADMIN) || access.includes(getRoleForDetector(name)));
```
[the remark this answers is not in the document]
xsalonx: applied

martinboulais (approved):

PR #1611, opened 2024-06-13
[O2B-1296] Fix data pass name usage in QC flags GRPC controller

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- NA

- Added fetching LHC Period in GRPC QC flags controller in order to build data pass name

Changes made to the database:
- NA


lib/server/controllers/gRPC/GRPCQcFlagController.js
```diff
@@ -26,13 +27,22 @@ class GRPCQcFlagController {
 
     // eslint-disable-next-line require-jsdoc
     async CreateForDataPass({ runNumber, detectorName, passName, flags }) {
+        const lhcPeriods = await LhcPeriodRepository.findOne({ include: [{ association: 'runs', where: { runNumber } }] });
```
martinboulais, line 30: NEVER use repositories in controller layer. You must have a function from the service for that.
What I would do though is to fetch the run and include the period from it?
You should have a service function `getFullDataPassNameUsingRunPeriod(partialPassName, runNumber)`

martinboulais (changes requested):

lib/server/controllers/gRPC/GRPCQcFlagController.js
xsalonx: applied

martinboulais (approved):

PR #1620, opened 2024-06-20
[O2B-1301] Disable QC for Not_for_physics runs

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Disabled QC for runs which are tags as `Not for physics` and marked such runs in RunPerLhcPeriod view

- NA

Changes made to the database:
- NA


lib/public/components/qcFlags/isRunSubjectToQc.js
```diff
@@ -0,0 +1,51 @@
...
+
+const QcDisabledReason = {
+    BAD_RUN_IN_ASYNC_QC_MESSAGE: `Quality of the run was changed to ${RunQualities.BAD} so it is no more subject to QC`,
+    NOT_FOR_PHYSICS_TAG: 'Run was tags as not for physics so it is no more subject to QC',
+};
+
+/**
+ * Set of names of tags which prevent run to be subject to QC
+ */
+const RUN_TAGS_DISABLING_QC = new Set(['Not for physics']);
```
martinboulais, line 25: No need for a set here, array is simpler, easier to read and (as far as I know) faster for few elements

lib/public/components/qcFlags/isRunSubjectToQc.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Get reason why a run is not subject to QC
+ *
+ * @param {Run} run a run
+ * @return {string} reason
+ */
+export const getRunNotSubjectToQcReason = ({ runQuality, tags }) => {
+    let reason = '';
+    if (runQuality === RunQualities.BAD) {
+        reason += QcDisabledReason.BAD_RUN_IN_ASYNC_QC_MESSAGE;
```
martinboulais, line 45: I see no reason to have `QcDisabledReason` instead of directly writing the string here. You are in a component, so it's fine to have text strings without constants.
String constants such as these ones goes a bit against the component system itself, if you need to factorize a string, you probably need to extract a component.

martinboulais (changes requested):

lib/public/components/qcFlags/isRunSubjectToQc.js
xsalonx: applied

lib/public/components/qcFlags/isRunSubjectToQc.js
xsalonx: applied

lib/public/components/qcFlags/isRunSubjectToQc.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @param {Run} run a run
+ * @return {string} reason
+ */
+export const getRunNotSubjectToQcReason = ({ runQuality, tags }) => {
+    let reason = '';
+    if (runQuality === RunQualities.BAD) {
+        reason += `Quality of the run was changed to ${RunQualities.BAD} so it is no more subject to QC`;
+    }
+    if (tags.find(({ text }) => RUN_TAGS_DISABLING_QC.has(text))) {
```
martinboulais, line 42: You forgot to update this has

lib/public/components/qcFlags/isRunSubjectToQc.js
```diff
@@ -0,0 +1,46 @@
...
+        reason += 'Run was tags as not for physics so it is no more subject to QC';
+    }
+    return reason;
```
martinboulais, line 45: I would rather return `reasons`, an array and then joint them as needed in caller. 
Because you might want to display them as list, or concatenate depending on the context.
Here, when run has both it displays `...to QCRun was tags...`

martinboulais (changes requested):

lib/public/components/qcFlags/isRunSubjectToQc.js
xsalonx: fixed

lib/public/components/qcFlags/isRunSubjectToQc.js
xsalonx: fixed

martinboulais (approved):

PR #1621, opened 2024-06-20
[O2B-1274] Add API for setting GAQ detectors 

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- NA

- Added API for setting GAQ detectors

Changes made to the database:
- NA


lib/server/controllers/qcFlag.controller.js
```diff
@@ -248,6 +248,36 @@ const getQcFlagsSummaryHandler = async (req, res) => {
     }
 };
 
+// eslint-disable-next-line valid-jsdoc
+/**
+ *
+ */
+const setGaqDetectors = async (req, res) => {
```
martinboulais, line 255: In my opinion, this belongs to DataPass controller not QC

lib/server/services/qualityControlFlag/QcFlagService.js
```diff
@@ -314,6 +316,117 @@ class QcFlagService {
...
+     * Set list of DPL detectors which contribute to GAQ (global aggregated quality)
+     * for given runs and data pass
+     *
+     * @param {object} scope of definition
+     * @param {number} scope.dataPassId data pass id
+     * @param {number[]} scope.runNumbers list of unique run numbers
+     * @param {number[]} scope.dplDetectorIds list of unique DPL detector ids
+     * @return {Promise<void>} promise
+     */
+    setGaqDetectors({ dataPassId, runNumbers, dplDetectorIds }) {
```
martinboulais, line 329: Do not seems to have anything to do with QC flags?

martinboulais, line 329: Why not passing actual parameters ? Scope make sense when you have varying parameters, I think it's not the case here. 
Also please try to be more descriptive in the parameters description

lib/server/controllers/qcFlag.controller.js
```diff
@@ -248,6 +248,36 @@ const getQcFlagsSummaryHandler = async (req, res) => {
...
+    const validatedDTO = await dtoValidator(
+        DtoFactory.bodyOnly(Joi.object({
+            dataPassId: Joi.number(),
+            runNumbers: Joi.array().items(Joi.number()).unique(),
```
martinboulais, line 259: From what you described, you will update GAQ only for runs one by one, no need to take an array here.

lib/server/services/qualityControlFlag/QcFlagService.js
```diff
@@ -314,6 +316,117 @@ class QcFlagService {
...
+            throw new BadParameterError(errorMessage);
+        }
+        const runsNotAssociatedToDataPass = validRunNumbers
+            .filter(({ dataPass }) => dataPass.length === 0)
+            .map(({ runNumber }) => runNumber);
+        if (runsNotAssociatedToDataPass.length > 0) {
+            const errorMessage = `No association between data pass with id ${dataPassId} and following runs: ${runsNotAssociatedToDataPass}`;
+            throw new BadParameterError(errorMessage);
+        }
+    }
```
martinboulais, line 376: I think there is no need for such complexity. You should simply check that if you count runs with this id(s) associated to data pass with this id, you're good (maybe you can even create a model on the association table? Not sure it's worth it)
As mentioned before, you update GAQ runs per runs, so this is easy 

This should probably go away from QC flags too

lib/server/services/qualityControlFlag/QcFlagService.js
```diff
@@ -314,6 +316,117 @@ class QcFlagService {
...
+
+    /**
+     * Validate association between runs and DPL detectors.
```
martinboulais, line 379: Validate is ambiguous here

lib/server/services/qualityControlFlag/QcFlagService.js
```diff
@@ -314,6 +316,117 @@ class QcFlagService {
...
+     * As there is no direct link between runs and DPL detectors,
+     *  association between run and corresponding Detector (with same name as DPL Detector) is checked
+     *
+     * @param {number[]} runNumbers list of run numbers
+     * @param {number[]} dplDetectorIds list of IDs of DPL detectors
+     * @return {Promise<void>} promise resolved once associations are validated
+     * @throws {BadParameterError} if some association does not exist
+     */
+    async _validateRunsAndDplDetectorsAssociations(runNumbers, dplDetectorIds) {
```
martinboulais, line 388: Same here, this can be simplified, if you handle runs one by one

martinboulais (changes requested):

lib/server/controllers/qcFlag.controller.js
xsalonx: From what I described for first feature in fronted , it is expected to do in batches from browser in the future and directly via API now as users use their own scripts, etc., so there is not place for such a simplification here

lib/server/services/qualityControlFlag/QcFlagService.js
xsalonx: https://github.com/AliceO2Group/Bookkeeping/pull/1621#discussion_r1673619269

lib/server/services/qualityControlFlag/QcFlagService.js
xsalonx: https://github.com/AliceO2Group/Bookkeeping/pull/1621#discussion_r1673619269

lib/server/services/qualityControlFlag/QcFlagService.js
xsalonx: simplified

lib/server/services/qualityControlFlag/QcFlagService.js
xsalonx: refactored

martinboulais (approved):

PR #1632, opened 2024-06-26
Remove default URLs from services config

xsalonx:
#### I DON'T have JIRA ticket
- [ ] tests are added

Notable changes for users:
- NA

- Removed default URLs from services config

Changes made to the database:
- NA


lib/config/services.js
```diff
@@ -23,22 +25,22 @@ const {
 exports.services = {
     enableHousekeeping: process.env?.ENABLE_HOUSEKEEPING ?? false,
     aliEcsGui: {
-        url: process.env?.ALI_ECS_GUI_URL || changeOriginPort(origin, 8080),
-        token: process.env?.ALI_ECS_GUI_TOKEN,
+        url: process.env?.ALI_ECS_GUI_URL || null,
```
martinboulais, line 28: You should use ?? instead of ||

martinboulais (changes requested):

lib/config/services.js
xsalonx: I'd say no, because if, in bash, you have statement like `export VAR=` then `process.env.VAR` evaluate it to empty string which remains if using `??`

martinboulais (approved):

PR #1660, opened 2024-07-18
Set proto User.externalId to explicit presence

teo:
Goes together with https://github.com/AliceO2Group/Control/pull/596 on the AliECS side so that an externalId set to 0 by the GUI to mean "Anonymous" is written to BK


graduta (approved):

martinboulais:
Hi @teo I think we are going in the wrong direction here.
externalId is **not** optional, as you have put in your PR, so using optional in proto shows a flaw somewhere in my opinion.
gRPC skips transmissions of **not-optional** values that have the default one, which make sense, and it's the role of the client to use default if it found nothing for a non-optional value. If the gRPC client implementation do not make it, I suppose it's the role of the adapter/controller/whatever to do the job and coalesce the value provided by the gRPC client.

teo:
@martinboulais the problem is a bit more subtle, unfortunately, and this was reported to me by @graduta. The GUI sends an externalId value `0` to AliECS if the user is "Anonymous", but if we keep it non-optional on the AliECS side then AliECS gets an empty value and BK gets the same. The problem is also that if the GUI queries back the user information it passed, it gets a different payload than the one it sent. So technically in this case externalId `0` is not the same as "no externalId provided". This is why I proposed to change it to `optional` in the BK protofile as well: if it's not, then BK doesn't get the same information that AliECS has.

martinboulais:
Ok if the externalId might actually be not-defined in this case it make sense to make it optional I agree, thanks for the clarification.

PR #1670, opened 2024-07-22
[O2B-1319] Merge detector tables - adjust run's detectors creation

xsalonx:
#### I have a JIRA ticket
- [x] branch and/or PR name(s) include(s) JIRA ID
- [ ] issue has "Fix version" assigned
- [ ] issue "Status" is set to "In review"
- [x] PR labels are selected

Notable changes for users:
- NA

Notable changes for developers:
- Adjusted run's detectors creation to work only with physical detectors

Changes made to the database:
- Added column type (ENUM) to detectors table

PR #1673, opened 2024-07-24
[O2B-1319] Merge detector tables -  adjust detectors provider

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- NA

- Refactored detectorsProvider to work with detector's type

Changes made to the database:
- NA


xsalonx:
To be merged after https://github.com/AliceO2Group/Bookkeeping/pull/1670

lib/server/services/dpl/getOrCreateDplDetector.js
```diff
@@ -11,12 +11,44 @@
...
  *
  * @param {string} name the name of the dpl detector to find/create
  * @return {Promise<SequelizeDplDetector>} the dpl detector found
  */
-exports.getOrCreateDplDetectorByName = (name) => DplDetectorRepository.findOneOrCreate({ name });
+exports.getOrCreateDplDetectorByName = async (name) => {
+    let dplDetector = await DplDetectorRepository.findOne({ where: { name } });
+    if (dplDetector) {
+        return dplDetector;
+    } else {
```
martinboulais, line 28: No need for this else, you can simply continue.

lib/server/services/dpl/getOrCreateDplDetector.js
```diff
@@ -11,12 +11,44 @@
...
+        /**
+         * Because of merging detectors and dpl_detector table,
+         * temporarly detectors and dpl_detectors with same name must also have same id
+         */
+        dplDetector = await DplDetectorRepository.insert({ name });
+        let detector = await DetectorRepository.findOne({ where: { name } });
+        if (!detector) {
+            detector = await DetectorRepository.insert({ name, type: DetectorType.QC });
+        }
+        const [[{ mid: maxDetectorId }]] = await sequelize.query('SELECT MAX(id) AS mid FROM detectors');
```
martinboulais, line 38: I don't understand why you want a common id? Id is auto-generated, you should avoid setting it manually in any of the two tables.
You'll update the columns of tables pointing to DPL detectors when you'll create the fk for it, and you'll simply use the name as criteria, not id (because current id in dpl detectors do not match the ones in detectors)

martinboulais (changes requested):

lib/database/migrations/20240724134539-migrate-dpl-detectors-to-detectors.js
```diff
@@ -0,0 +1,49 @@
...
+            WHERE name = '' AND name NOT IN (SELECT name from detectors)
+        `, { transaction });
+    }),
+
+    down: async (queryInterface) => queryInterface.sequelize.transaction(async (transaction) => {
+        await queryInterface.sequelize.query(`
+        DELETE FROM detectors
+            WHERE type NOT IN (${DATA_TAKING_DETECTOR_TYPES.map((type) => `'${type}'`)})
+        `, { transaction });
+    }),
```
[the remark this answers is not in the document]
xsalonx: no,  because nothing is deleted from dpl_detectors

lib/public/services/detectors/detectorsProvider.js
```diff
@@ -42,6 +43,33 @@ export class DetectorsProvider extends Observable {
...
             .build();
+
+        this._dataTaking$ = ObservableData.builder()
+            .source(this._all$)
+            .apply((remoteDetectors) => remoteDetectors.apply({
+                Success: (detectors) => detectors.filter(({ type }) => DATA_TAKING_DETECTOR_TYPES.includes(type)),
+            }))
+            .build();
+
+        this._qc$ = ObservableData.builder()
```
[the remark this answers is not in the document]
xsalonx: ok

lib/server/services/dpl/getOrCreateDplDetector.js
xsalonx: because: on frontend, only `detectorsProvider` remains, which will use `detectors` (not `dpl_detectors`) so e.g. QcFlag views will use identifiers of `detectors`, but still there are sql/sequelize views which use `dpl_detectors` and some FK constraints, in the migration you can see that all `detectors` have IDs copied from `dpl_detectors`, but in case something happened and next step cannot be taken soon, and still there is `dpl_detectors` table IDs in in both tables should still be the same. In order to avoid doing everything at once, this temporary solution allow this, also setting auto-generated id manually doesn't break anything

lib/server/services/dpl/getOrCreateDplDetector.js
xsalonx: applied

lib/server/services/dpl/getOrCreateDplDetector.js
martinboulais: I did not notice but in the migration you update the ids of detectors. This is not correct, because all the other tables that points to this will have wrong links now (for example: runs). You must not do that. In general, updating id is a very bad idea.

This probably mean that for now you can't delete the dplDetectors provider

xsalonx: yes I can updated, because there is foriegn key constraint (onUpdate: CASCADE), and only reference to detectors table is in run_detectors table

PR #1681, opened 2024-07-25
[O2B-1222] Stable beam only toggle on lhc fills overview page

AnoeBanoe:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Toggle button on lhc overview page to show either all lhc fills or only ones with stable beam


lib/public/views/LhcFills/Overview/LhcFillsOverviewModel.js
```diff
@@ -29,6 +30,8 @@ export class LhcFillsOverviewModel extends OverviewPageModel {
         super();
 
         this._stableBeamsOnly = stableBeamsOnly;
+
+        this._debouncedLoad = debounce(this.load.bind(this), 200);
```
martinboulais, line 34: I think that the debounce is not useful, it might make sense when user fill in text but not the way we do it.

lib/public/views/LhcFills/Overview/index.js
```diff
@@ -18,11 +18,26 @@ import { lhcFillsActiveColumns } from '../ActiveColumns/lhcFillsActiveColumns.js
...
+ * @returns {Component} the toggle switch
+ */
+export const toggleStableBeamOnlyFilter = (lhcFillsOverviewModel) => {
+    const isStableBeamsOnly = lhcFillsOverviewModel.isStableBeamsOnly();
+    const onChange = isStableBeamsOnly
+        ? () => lhcFillsOverviewModel.setStableBeamsFilter(false)
+        : () => lhcFillsOverviewModel.setStableBeamsFilter(true);
+    return switchInput(isStableBeamsOnly, onChange, { labelAfter: 'STABLE BEAM ONLY' });
+};
+
```
martinboulais, line 40: Here the onchange should not depend on external context, it should simply take as parameter the new state and set the filter accordingly

martinboulais (changes requested):

lib/public/views/LhcFills/Overview/index.js
AnoeBanoe: Done

lib/public/views/LhcFills/Overview/LhcFillsOverviewModel.js
AnoeBanoe: Done

lib/public/views/LhcFills/Overview/index.js
martinboulais: Are you sure?

martinboulais (changes requested):

martinboulais (approved):

PR #1677, opened 2024-07-25
Uniformize enums

martinboulais:
#### I have a JIRA ticket
- [X] branch and/or PR name(s) include(s) JIRA ID
- [X] issue has "Fix version" assigned
- [X] issue "Status" is set to "In review"
- [X] PR labels are selected

Notable changes for developers:
- Uniformized enums: all are frozen objects in dedicated files, named in PascalCase with properties in CAPITAL_SNAKE_CASE

PR #1687, opened 2024-07-30
[O2B-1325] GAQ summary display

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Added GAQ summary to Runs Per Data Pass view

- NA

Changes made to the database:
- NA


xsalonx:
To be merged after https://github.com/AliceO2Group/Bookkeeping/pull/1683

lib/public/views/Runs/ActiveColumns/getQcSummaryDisplay.js
```diff
@@ -0,0 +1,111 @@
...
+
+    const missingVerificationMessage = `Missing ${missingVerificationsCount} verification${missingVerificationsCount > 1 ? 's' : ''}`;
+    const missingVerificationDisplay = missingVerificationsCount
+        ? h(
+            '.d-inline-block.va-t-bottom',
+            tooltip(h('.f7', iconWarning()), missingVerificationMessage),
+        )
+        : null;
+
+    const notBadPercentageFormat = h('span', formatFloat((1 - badEffectiveRunCoverage) * 100, { precision: 0 }));
```
graduta, line 44: Will this `notBadPercentageFormat` be reused in other views? In that case perhaps it is better to extract it in a component of its own in the `format` directory?
Moreover, if you plan to display a percentage why not use `formatPercentage` instead of `formatFloat`?

lib/public/views/Runs/ActiveColumns/getQcSummaryDisplay.js
```diff
@@ -0,0 +1,111 @@
...
+ */
+export const getQcSummaryDisplay = (summary) => {
+    const {
+        badEffectiveRunCoverage,
+        explicitlyNotBadEffectiveRunCoverage,
+        missingVerificationsCount,
+        mcReproducible,
+    } = summary;
+
+    const missingVerificationMessage = `Missing ${missingVerificationsCount} verification${missingVerificationsCount > 1 ? 's' : ''}`;
```
graduta, line 36: As this `const` is only used once in the ternary condition and only if there are missing verifications count, I would say the message can be declared directed in the `h()` method

lib/public/views/Runs/ActiveColumns/getQcSummaryDisplay.js
```diff
@@ -0,0 +1,111 @@
...
+        notBadPercentageFormat,
+        missingVerificationDisplay,
+    ];
+
+    /**
+     * Format QC summary with color
+     * @param {string|Component} content content
+     * @param {string} [configuration.color] the color to look for contrast,
+     * in hexa RGB or CSS var() function over variable name expression: e.g. "var(--color-nice)"
+     * @returns {Component} QC summary display
```
graduta, line 56: You are using `return` above and `returns` here.
Please keep to `return` usage

lib/public/views/Runs/ActiveColumns/getQcSummaryDisplay.js
```diff
@@ -0,0 +1,111 @@
...
+        qcSummaryDisplay = getQcSummaryDisplayWithColor(
+            [
+                h('.d-inline-block', tooltip(
+                    iconBolt(),
+                    'Run start or stop is missing and time-based flag was assigned, coverage cannot be calculated',
+                )),
+                missingVerificationDisplay,
+            ],
+            {
+                color: QC_SUMMARY_COLORS.INCALCULABLE_COVERAGE,
```
graduta, line 78: The convention used in Bookkeeping for enums is that the key is in upper camelcase (`QcSummaryColors`) and the keys within the object are capital snake case (INCALCULABLE_COVERAGE).
Please update the enum accordingly

graduta: For reference: https://github.com/AliceO2Group/Bookkeeping/pull/1677

lib/public/views/Runs/ActiveColumns/runDetectorsAsyncQcActiveColumns.js
```diff
@@ -35,86 +31,7 @@ const FULL_COVERAGE = 1;
...
-        qcSummaryDisplay = getQcSummaryDisplayWithColor(
-            nonMcReproducibleQcSummaryContent,
-            {
-                color: badEffectiveRunCoverage === FULL_COVERAGE
-                    ? QC_SUMMARY_COLORS.ALL_BAD
-                    : QC_SUMMARY_COLORS.PARTIALLY_BAD,
-            },
-        );
-    }
+    let qcSummaryDisplay = getQcSummaryDisplay(qcSummary[runNumber][dplDetectorId]);
```
graduta, line 34: I think it is best to not declare and generate the content of the summary display as a let here and then reassign based on a ternary condition.
Instead, only call the `getQcSummaryDisplay` in the ternary condition if it meets the criteria (L38)

graduta (changes requested):

lib/public/views/Runs/ActiveColumns/getQcSummaryDisplay.js
xsalonx: (in other components), for now, it won't be used outside this component.
`formatPercentage` adds `%` at the end, which is not desired here

lib/public/views/Runs/ActiveColumns/getQcSummaryDisplay.js
xsalonx: fixed

lib/public/views/Runs/ActiveColumns/getQcSummaryDisplay.js
xsalonx: applied

lib/public/views/Runs/ActiveColumns/getQcSummaryDisplay.js
xsalonx: applied

lib/public/views/Runs/ActiveColumns/runDetectorsAsyncQcActiveColumns.js
xsalonx: I would say, it will decrease readability

graduta: Why do you think it will descrease readability? You are already applying the `tooltip` in the ternary already rather than having it as a `const`. 

Would you say that this is better 
```js
let qcSummaryDisplay = getQcSummaryDisplay(qcSummary[runNumber][dplDetectorId]);
 qcSummaryDisplay = isRunNotSubjectToQc(run)
        ? tooltip(qcSummaryDisplay, getRunNotSubjectToQcReason(run))
        : qcSummaryDisplay;
```
than this?
```js
const qcSummaryDisplay = isRunNotSubjectToQc(run)
        ? tooltip(qcSummaryDisplay, getRunNotSubjectToQcReason(run))
        : getQcSummaryDisplay(qcSummary[runNumber][dplDetectorId]);
```

To me is the exact opposite for readability + the benefit of efficiency

graduta (approved):

PR #1690, opened 2024-07-31
[O2B-1327] User interface for setting GAQ detectors

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Added form for choosing GAQ detectors

- Added 'gaq' role to (dev) UserRoleSelectionModel

Changes made to the database:
- NA


xsalonx:
To be merged after https://github.com/AliceO2Group/Bookkeeping/pull/1693

lib/public/components/common/selection/picker/picker.js
```diff
@@ -94,16 +98,23 @@ export const picker = (pickerOptions, pickerModel, configuration) => {
...
-            type: 'search',
-            placeHolder: '🔍 Search for tags',
-            value: pickerModel.searchInputContent,
-            oninput: (e) => {
-                pickerModel.searchInputContent = e.target.value;
-            },
-        }),
-        applyAttributes(checkboxes.length ? checkboxes : h('em', 'No options')),
-    ];
+    if (searchEnabled) {
```
graduta, line 101: Is it not better to simply add the `searchEnabled` check in front of the `h` element of the search?
Or event better extract the `h(input [...])` into its own component and then use:
```js
searchEnabled && searchInputElement()
```

lib/public/components/detector/PhysicalDetectorPickerModel.js
```diff
@@ -0,0 +1,55 @@
...
+import { PickerModel } from '../common/selection/picker/PickerModel.js';
+import { badge } from '../common/badge.js';
+
+/**
+ * Map detector to select options with the detector badge as label
+ *
+ * @param {Detector} detector detector
+ * @return {SelectOption} select option
+ */
+export const detectorToOption = ({ id, name }) => ({
```
graduta, line 25: As this will be used for options of `h()` element, I think the `id` and `key` should be used as it is an iteration

graduta (changes requested):
There are some improvements that could and should be done on the UI:
- given that you already load the run data, thus detectors part of that run, should the modal not include only those detectors? Otherwise we always display a DB error message. It will be easier for user to know which eligible detectors there are to select from;
- please have a look at display flex and justify-content so that checkboxes are aligned vertically with the label.

lib/public/components/detector/PhysicalDetectorPickerModel.js
xsalonx: it's responsibility of picker , not picker model

graduta (approved):

graduta (approved):

PR #1696, opened 2024-08-02
Replace legacy docker-compose

xsalonx:
#### I DON'T have JIRA ticket
- [ ] tests are added

Notable changes for users:
- NA

- Replaces `docker-compose` command by `docker compose`
- Updates files name as per docker requirements: https://docs.docker.com/compose/project-name/#set-a-project-name 

Changes made to the database:
- NA


.github/workflows/bookkeeping.yml
```diff
@@ -17,20 +17,20 @@ jobs:
       matrix:
         test_type: [unit,
                     api,
-                    lhcPeriods,
```
martinboulais, line 20: Why this change? I think it follows our js file name convention, even if I prefer your way I think we should keep the original.

martinboulais (dismissed):

.github/workflows/bookkeeping.yml
xsalonx: I thought it resolves issue with hanged jobs: 
image
, but apparently no

xsalonx: but following instruction is `"COMPOSE_PROJECT_NAME=${{ matrix.test_type }}_app"` `test_type` cannot has upercase letters

graduta: @xsalonx is referring to :
https://docs.docker.com/compose/project-name/#set-a-project-name

graduta (approved):

.github/workflows/bookkeeping.yml
graduta: The `hanging jobs` are due to the fact that GH PR rules are waiting for those jobs names but because you are changing them, they are never being created.
I will update the rules accordingly

PR #1749, opened 2024-09-03
Start docker compose in detached mode

martinboulais:
#### I DON'T have JIRA ticket
- [X] explain what this PR does
- [X] if it is a new feature, explain how you plan to use it
- [X] documentation was updated or added

- Docker compose stack is now run in detached mode: this allow user to start docker stack from console without having to keep this console instance opened.


martinboulais:
Following discussion with colleagues, will keep the non-detached mode.

PR #1774, opened 2024-10-03
[O2B-1386] Check if clipboard is available

mariscalromeroalejandro:
#### I DON'T have JIRA ticket
- [ ] explain what this PR does
- [ ] tests are added

#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- Users can only press "copy to..." button when the clipboard is available. Otherwise, they will see a message explaining the reason.

-

Changes made to the database:
-

### DIFF
--- a/lib/public/components/common/selection/infoLoggerButtonGroup/CopyToClipboardComponent.js
+++ b/lib/public/components/common/selection/infoLoggerButtonGroup/CopyToClipboardComponent.js
@@ -42,6 +42,56 @@ export class CopyToClipboardComponent extends StatefulComponent {
     }
 
     /**
+     * Checks if the clipboard is available and returns an object with availability status and a message.
+     *
+     * @returns {{ available: boolean, message: string }}
+     * - `available`: True if the clipboard is available; false otherwise.
+     * - `message`: A reason if the clipboard is not available, otherwise an empty string.
+     */
+    isClipboardAvailable() {
+        if (!this.isContextSecure()) {
+            return { available: false, message: 'Clipboard not available in a non-secure context.' };
+        }
+
+        if (!this.isClipboardSupported()) {
+            return { available: false, message: 'Clipboard API is not supported in this browser.',
+            };
+        }
+
+        if (this.isWindowEmbedded()) {
+            return { available: false, message: 'Clipboard access is restricted in iframes.' };
+        }
+        return { available: true, message: '' };
+    }
+
+    /**
+     * Checks if context is secure (HTTPS)
+     *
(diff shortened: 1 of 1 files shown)


lib/public/components/common/selection/infoLoggerButtonGroup/CopyToClipboardComponent.js
```diff
@@ -41,6 +41,56 @@ export class CopyToClipboardComponent extends StatefulComponent {
         this.notify();
     }
 
+    /**
+     * Checks if the clipboard is available and returns an object with availability status and a message.
+     *
+     * @returns {{ available: boolean, message: string }}
+     * - `available`: True if the clipboard is available; false otherwise.
+     * - `message`: A reason if the clipboard is not available, otherwise an empty string.
```
martinboulais, line 49: If message do not apply (meaning: it makes no sense to have a message, in the case the clipboard is available) you should rather use null or undefined (basically return `{ available: true }`).

Another possibility is to have a function like `checkClipboardAvailability` that throws an error if one of the check fails, and you use the error message as message.

lib/public/components/common/selection/infoLoggerButtonGroup/CopyToClipboardComponent.js
```diff
@@ -59,6 +110,8 @@ export class CopyToClipboardComponent extends StatefulComponent {
             {
                 id: `copy-${id}`,
                 onclick: () => this.copyToClipboard(clipboardTargetValue),
+                disabled: !available,
+                title: `${message}`,
```
martinboulais, line 114: If message is already a string, no need for ``

martinboulais (changes requested):

lib/public/components/common/selection/infoLoggerButtonGroup/CopyToClipboardComponent.js
```diff
@@ -48,20 +48,19 @@ export class CopyToClipboardComponent extends StatefulComponent {
...
-            };
+            throw new Error('Clipboard API is not supported in this browser.');
         }
 
         if (this.isWindowEmbedded()) {
-            return { available: false, message: 'Clipboard access is restricted in iframes.' };
+            throw new Error('Clipboard access is restricted in iframes.');
         }
-        return { available: true, message: '' };
+        return { available: true };
```
martinboulais, line 63: You don't need to return anything in check, you simply run it and wait for it not to throw.

martinboulais (dismissed):

martinboulais (approved):

PR #1786, opened 2024-10-27
[O2B-1390] Fix grey QC summary

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Prevented the QC summary color from changing to gray after more than one verification 

- Added separate request to fetch non-verified flags when calculating QC summary

Changes made to the database:
- NA

### DIFF
--- a/lib/server/services/qualityControlFlag/QcFlagService.js
+++ b/lib/server/services/qualityControlFlag/QcFlagService.js
@@ -321,11 +321,14 @@ class QcFlagService {
                     mcReproducible: Boolean(summaryDb.get('mcReproducible')),
                 }));
 
-        const s = new Set(runDetectorSummaryList.flatMap(({ flagIds }) => flagIds));
-        const verifiedFlagsIds = new Set((await QcFlagRepository.findAll({
+        const allFlagsIds = new Set(runDetectorSummaryList.flatMap(({ flagIds }) => flagIds));
+        const notVerifiedFlagsIds = new Set((await QcFlagRepository.findAll({
             attributes: ['id'],
-            include: [{ association: 'verifications', required: true, attributes: [] }],
-            where: { id: { [Op.in]: [...s] } },
+            include: [{ association: 'verifications', required: false, attributes: [] }],
+            where: {
+                id: { [Op.in]: [...allFlagsIds] },
+                '$verifications.id$': { [Op.is]: null },
+            },
         })).map(({ id }) => id));
 
         const summary = {};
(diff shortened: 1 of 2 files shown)


lib/server/services/qualityControlFlag/QcFlagService.js
```diff
@@ -318,19 +317,30 @@ class QcFlagService {
                     detectorId: summaryDb.detectorId,
                     effectiveRunCoverage: parseFloat(summaryDb.get('effectiveRunCoverage'), 10) || null,
                     bad: Boolean(summaryDb.get('bad')),
-                    missingVerificationsCount: parseInt(summaryDb.get('missingVerificationsCount'), 10),
+                    flagIds: (summaryDb.get('flagIds')?.split(',') ?? []).map((id) => parseInt(id, 10)),
                     mcReproducible: Boolean(summaryDb.get('mcReproducible')),
                 }));
 
+        const allFlagsIds = new Set(runDetectorSummaryList.flatMap(({ flagIds }) => flagIds));
+        const notVerifiedFlagsIds = new Set((await QcFlagRepository.findAll({
```
martinboulais, line 325: Can't you use a group by instead of a set?

martinboulais (changes requested):
Could you detail a bit why there were this bug and how this solves it?

lib/server/services/qualityControlFlag/QcFlagService.js
xsalonx: I am not using Set because there are duplicated, but to have faster filtering
```
const missingVerificationsCount = flagIds.filter((id) => notVerifiedFlagsIds.has(id)).length;
```

xsalonx:
In current design on main, QcFlag verification are included in the query for sake of counting non-verified flags:
```
attributes: [
...
                [
                    sequelize.literal('COUNT( DISTINCT `QcFlag`.id ) - COUNT( DISTINCT `verifications`.flag_id )'),
                    'missingVerificationsCount',
                ],
```
so sequelize adds `LEFT OUT JOIN quality_control_flag_verification as ...` which (in combination with join with effective periods) cause duplication of effective periods entries in the SUM calculating `effectiveRunCoverage`, afaik It can be help neither with additional `group by` or `distinct`. It can be help by subquery, so  `effectiveRunCoverage` is calculated based on sub-select of unique effective periods. I this PR I proposed to do differently, by counting missing verification separately.

martinboulais (approved):

PR #1805, opened 2024-12-02
[O2B-1139] Refresh tags list when a tag is updated/created

hattinf:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
-tag colours will now update when looking at run details

-

Changes made to the database:
-


lib/public/views/Runs/Details/RunDetailsModel.js
```diff
@@ -76,6 +76,9 @@ export class RunDetailsModel extends Observable {
 
             this._fetchOneRun();
             this._fetchReasonTypes();
+        } else {
+            this._runNumber = runNumber;
+            this._fetchOneRun();
```
martinboulais, line 81: When you end up with one line being the same in if and in else, often you should simply move it out of the if. The `_fetchReasonTypes` and `_fetchOneRun` are both asynchronous and you do not care about the order, so you can swap them. Same goes for the `this._runNumber = runNumber`. Then you see that you end both the if and else with the same lines (setting `this._runNumber` and `_fetchOneRun`) so you can move it after the if/else.

Another point, you have not put `this.runDetails = RemoteData.notAsked` probably because you see that this is overriden right after in `this._fetchOneRun`, I suggest that you remove it also from the if (same goes for the `this._eorReasonTypes = RemoteData.notAsked();`

martinboulais (commented):
Please add one test that covers the bug first (and expect it to fail), then implement the fix and see if the test pass

test/public/runs/detail.test.js
```diff
@@ -527,4 +553,29 @@ module.exports = () => {
...
     });
+
+    it('should display correct tag styling after updating in tag overview', async () => {
+        let badges;
+        // Fetch the run data before update of tag
+        await goToRunDetails(page, 106);
+
+        badges = await getBadges();
+
+        expect(badges[0].backgroundColor === 'rgb(255, 0, 0)').to.be.false;
```
martinboulais, line 564: If you always use `badges[0].backgroundColor`, you should have a function that simply return the background color of the first badge (YAGNI principle, the simplest is often the best). This function should be named accordingly.

Also, tests should be precises. int this case, you already know from the seeders what is the color of the badge, you should expect this color to be true (and not the badge to be `not red`)

test/public/runs/detail.test.js
```diff
@@ -527,4 +553,29 @@ module.exports = () => {
...
+
+        const updatedTag = {
+            color: '#ff0000', // Red color
+        };
+
+        // Update the tag via API request
+        await request(server)
+            .put('/api/tags/1?token=admin')
+            .send(updatedTag)
+            .expect(201);
```
martinboulais, line 574: Yes that works, but I think you can make it more maintainable by directly calling the UpdateTagUseCase to avoid relying on the API.

martinboulais (changes requested):

test/public/runs/detail.test.js
```diff
@@ -64,6 +66,30 @@ module.exports = () => {
     let url;
     let createdRunId;
 
+    /**
+     *  Retrieve the badge classes and styles
+     *
+     *  @return {Promise<void>} A promise that resolves when the badge classes and styles are retrieved
```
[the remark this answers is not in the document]
hattinf: Done

test/public/runs/detail.test.js
```diff
@@ -64,6 +66,30 @@ module.exports = () => {
...
+    const getBadges = async () => {
+    // Check if the tag is updated
+        const tagsBadgeClassesSelector = '#Run-tags .badge';
+        // Wait for badge elements to appear
+        await page.waitForSelector(tagsBadgeClassesSelector);
+
+        // Evaluate and check for inline background color
+        const badgesWithStyles = await page.$$eval(
+            tagsBadgeClassesSelector,
+            (badges) => badges.map((badge) => ({
```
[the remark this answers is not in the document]
hattinf: Done

test/public/runs/detail.test.js
```diff
@@ -64,6 +66,30 @@ module.exports = () => {
...
     let createdRunId;
 
+    /**
+     *  Retrieve the badge classes and styles
+     *
+     *  @return {Promise<void>} A promise that resolves when the badge classes and styles are retrieved
+     */
+    const getBadges = async () => {
+    // Check if the tag is updated
+        const tagsBadgeClassesSelector = '#Run-tags .badge';
```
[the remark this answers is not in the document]
hattinf: Done

test/public/runs/detail.test.js
hattinf: Done

martinboulais (changes requested):

test/public/runs/detail.test.js
```diff
@@ -527,4 +529,58 @@ module.exports = () => {
...
+                tagsBadgeClassesSelector,
+                (badges) => badges.map((badge) => ({
+                    className: badge.className,
+                    backgroundColor: badge.style.backgroundColor,
+                })),
+            );
+        };
+
+        let badges;
+        const expectedBgColorBefore = 'badge b1 b-gray-light bg-gray-light';
```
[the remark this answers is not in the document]
hattinf: done

test/public/runs/detail.test.js
```diff
@@ -505,4 +506,57 @@ module.exports = () => {
...
+            // Check if the tag is updated
+            const tagsBadgeClassesSelector = '#Run-tags .badge';
+            // Wait for badge elements to appear
+            await page.waitForSelector(tagsBadgeClassesSelector);
+            // Evaluate and check for inline background color
+            return await page.$$eval(
+                tagsBadgeClassesSelector,
+                (badges) => badges.map((badge) => ({
+                    backgroundColor: badge.style.backgroundColor,
+                })),
```
martinboulais, line 526:
```suggestion
                (badges) => badges.map((badge) => badge.style.backgroundColor),
```

test/public/runs/detail.test.js
```diff
@@ -505,4 +506,57 @@ module.exports = () => {
...
+        let badges;
+        const expectedBgColorBefore = 'rgb(238, 238, 238)'; //Gray
+        const expectedBgColorAfter = 'rgb(255, 0, 0)'; //Red
+
+        // Fetch the run data before update of tag
+        await goToRunDetails(page, 106);
+
+        badges = await getRunTagsBadges();
+
+        expect(badges[0].backgroundColor == expectedBgColorBefore).to.be.true;
```
martinboulais, line 539:
```suggestion
        expect(badges[0]).to.equal(expectedBgColorBefore);
```

Doing so, you have explicit error message when values are not equal, something like `expected rgb(...) to equal rgb(...)`
When you expect a value of an array, putting an expect on the array length is often a good idea

test/public/runs/detail.test.js
```diff
@@ -505,4 +506,57 @@ module.exports = () => {
...
+                name: 'John Doe',
+            },
+        };
+        await new UpdateTagUseCase()
+            .execute(updateTagDto);
+
+        await goToRunDetails(page, 106);
+        badges = await getRunTagsBadges();
+
+        expect(badges[0].backgroundColor == expectedBgColorAfter).to.be.true;
```
martinboulais, line 560:
```suggestion
        expect(badges[0]).to.equal(expectedBgColorAfter);
```

martinboulais (changes requested):

test/public/runs/detail.test.js
hattinf: Done

test/public/runs/detail.test.js
hattinf: Done

martinboulais (dismissed):

PR #1810, opened 2024-12-05
[O2B-719] Add range based filtering for run number

hattinf:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
-runNumber filter now accepts ranges as well

-

Changes made to the database:
-


lib/usecases/run/GetAllRunsUseCase.js
```diff
@@ -78,16 +78,33 @@ class GetAllRunsUseCase {
...
-                if (runNumberList.length) {
-                    if (runNumberList.length > 1) {
-                        // Multiple run numbers.
-                        filteringQueryBuilder.where('runNumber').oneOf(...runNumberList);
+                const allRuns = new Set();
+
+                runNumberList.forEach((runNumber) => {
+                    if (runNumber.includes('-')) {
+                        const [start, end] = runNumber.split('-').map((n) => parseInt(n, 10));
+                        if (!Number.isNaN(start) && !Number.isNaN(end) && start <= end) {
```
martinboulais, line 89: The last test is actually already covered by the loop

lib/usecases/run/GetAllRunsUseCase.js
```diff
@@ -78,16 +78,33 @@ class GetAllRunsUseCase {
 
             if (runNumbers) {
                 const runNumberList = runNumbers.split(SEARCH_ITEMS_SEPARATOR)
```
martinboulais, line 80: This should be renamed to `runNumberCriteria`, and your list `allRuns` should be something like `runNumberSet` (allRuns would mean runs entities)

martinboulais (changes requested):

lib/usecases/run/GetAllRunsUseCase.js
hattinf: What do you mean by last test?

martinboulais: I mean `start <= end` is included in the loop. You will not enter the loop if start > end

hattinf: done

lib/usecases/run/GetAllRunsUseCase.js
hattinf: Done

martinboulais:
The idea is good, but there is a major flaw: when using a very big range (with prod data) the app simply crash because the list of run numbers is too big.
A possible solution is to limit the range, for example to 100 (please read Joi documentation to put it in the DTO).
The best solution would be to use min and max SQL function for run, but I am not sure we can properly do that with the current query builder.

lib/domain/dtos/filters/RunFilterDto.js
```diff
@@ -29,8 +29,39 @@ const EorReasonFilterDto = Joi.object({
...
+    for (const runNumber of runNumbers) {
+        if (runNumber.includes('-')) {
+            const [start, end] = runNumber.split('-').map((n) => parseInt(n, 10));
+            if (Number.isNaN(start) || Number.isNaN(end) || start > end) {
+                return helpers.error('any.invalid', { message: `Invalid range: ${runNumber}` });
+            }
+            const rangeSize = end - start + 1;
+
+            if (rangeSize > MAX_RANGE_SIZE) {
+                return helpers.error('any.invalid', { message: `Range exceeds max size of 100 runs: ${runNumber}` });
```
martinboulais, line 53: You should use the constant also in the error message

martinboulais (changes requested):

lib/domain/dtos/filters/RunFilterDto.js
martinboulais: I meant instead of the hard coded 100, now I think you put twice the range

hattinf: Done, haha my bad thought you meant rangeSize

martinboulais (dismissed):

martinboulais (approved):

PR #1814, opened 2024-12-11
[O2B-1412] Removed retries limit to kafka connection

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- N/A

- Removed the max number of retries for kafka connection

Changes made to the database:
- N/A


graduta (commented):
@martinboulais do you happen to know if we can provide an interval for these retries?
Even better, an incremental interval?

martinboulais:
> @martinboulais do you happen to know if we can provide an interval for these retries? Even better, an incremental interval?

@graduta as far as I understand, the default behavior is actually exponential increment: https://kafka.js.org/docs/configuration#default-retry

graduta (approved):

PR #1820, opened 2024-12-17
[O2B-1056] bookkeeping error page

hattinf:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
-redirects to error page when page parameter is unknown

-

Changes made to the database:
-


lib/public/views/Error/ErrorModel.js
```diff
@@ -0,0 +1,53 @@
...
+        this.error = {
+            code: 'Unknown Error',
+            codeDescription: 'Something unexpected happened.',
+            message: 'Please try again later.',
+        };
+    }
+
+    /**
+     * Sets the error object for the model
+     * @param {Object} error The error object to set, must contain a code, codeDescription and message
```
martinboulais, line 35: Please use a precise object description instead of simply suggesting it in comment. Because you use it in different places, I suppose you should create a typedef.

lib/public/Model.js
```diff
@@ -13,31 +13,32 @@
...
-import { UserRoleSelectionModel } from './models/UserRoleSelectionModel.js';
-import { ObservableData } from './utilities/ObservableData.js';
-import { BkpRoles } from './domain/enums/BkpRoles.js';
-import { getRoleForDetector } from './utilities/getRoleForDetector.js';
-import { detectorsProvider } from './services/detectors/detectorsProvider.js';
+import RunTypeModel from './views/RunTypes/RunTypeModel.js';
+import { RunsModel } from './views/Runs/RunsModel.js';
+import { SimulationPassesModel } from './views/SimulationPasses/SimulationPassesModel.js';
+import { StatisticsPageModel } from './views/Statistics/StatisticsPageModel.js';
+import Tags from './views/Tags/Tags.js';
```
martinboulais, line 40: You should try to avoid as much as possible this kind of huge changes, for example reordering imports. It adds a lot of noise for review, and someone might do the other way around in a future commit, and this is an endless cycle of changes.

lib/public/views/Error/index.js
```diff
@@ -0,0 +1,41 @@
...
+                src: 'assets/alice.png',
+                alt: 'Alice logo',
+                style: 'width: 200px',
+            }),
+            h('h2', 'Oops! Something went wrong.'),
+            h('h3', `${code} - ${codeDescription}`),
+            h('.f5', message),
+            h('button.btn.btn-primary', {
+                onclick: () => {
+                    model.router.go('?page=home');
```
martinboulais, line 34: In my opinion, we should avoid as much as possible dependencies inside components. The ideal would be for it to depend only on a very specific submodel, in this case only on `model.errorModel`. Using router is an external dependency that should be avoided outside of models.
Instead, you should try to have a look at "frontLink" component, that hides this dependency for you (and puts it in one single place). You might style it as a button if you like, but in my personal opinion, a link should look like a link (I am fine with the button if you prefer it).

lib/public/views/Error/index.js
```diff
@@ -0,0 +1,41 @@
...
+                } }, [
+                iconHome(),
+                ' Go to Home Page',
```
martinboulais, line 37: You should use flex container and flex-gap instead of putting a space in your text

test/public/error/error.test.js
```diff
@@ -0,0 +1,62 @@
...
+    let page;
+    let browser;
+
+    before(async () => {
+        [page, browser] = await defaultBefore(page, browser);
+        await page.setViewport({
+            width: 700,
+            height: 720,
+            deviceScaleFactor: 1,
+        });
```
martinboulais, line 30: Do you need to set this?

test/public/error/error.test.js
```diff
@@ -0,0 +1,62 @@
...
+        // We expect the page to return the correct status code, making sure the server is running properly
+        expect(response.status()).to.equal(200);
+
+        // We expect the page to return the correct title, making sure there isn't another server running on this port
+        const title = await page.title();
+        expect(title).to.equal('AliceO2 Bookkeeping');
+    });
+
+    it('shows the error message', async () => {
+        await goToPage(page, 'error');
```
martinboulais, line 50: You can consider that one suite of frontend test is a continuous test, basically you can make them dependent one of the other. In this test for example you can suppose that you are already in the right page, and simply do your tests. Same for the next ones, and this way  we avoid navigation.

martinboulais (changes requested):

lib/public/views/Error/index.js
hattinf: done

test/public/error/error.test.js
hattinf: Removed it, because it was infact redundant

test/public/error/error.test.js
hattinf: Removed them

lib/public/Model.js
martinboulais: Then could you please revert it? Using git revert for example

lib/public/views/Error/index.js
martinboulais: The icon was a good idea, you should just use flex to display it properly with the text

martinboulais (changes requested):

lib/public/views/Error/ErrorModel.js
```diff
@@ -17,6 +17,14 @@ import { Observable } from '/js/src/index.js';
  * Model representing handlers for errorPage.js
  */
 export class ErrorModel extends Observable {
+    /**
+     * The Error object.
+     * @typedef {(Object)} Error
```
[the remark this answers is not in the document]
hattinf: Done

lib/public/views/Error/index.js
hattinf: Added it back, thought it looked weird if we were going for a link instead of a button, but now that I see it, it looks fine

martinboulais (approved):

PR #1834, opened 2025-01-15
[O2B-1290] multi run flagging backend

Jaden505:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
-

-

Changes made to the database:
-


lib/server/services/qualityControlFlag/QcFlagService.js
```diff
@@ -432,7 +432,7 @@ class QcFlagService {
                     to: newInstance.to,
                 });
 
-                createdFlags.push(qcFlagAdapter.toEntity(createdFlag));
+                createdFlags.push({ from, to, comment, runNumber, dplDetectorId: detector.id, flagTypeId, id: newInstance.id });
```
martinboulais, line 435: Why not using the adapter?

martinboulais (changes requested):

lib/server/services/qualityControlFlag/QcFlagService.js
Jaden505: done

lib/server/controllers/qcFlag.controller.js
```diff
@@ -147,49 +147,66 @@ const listSynchronousQcFlagsHandler = async (req, res) => {
...
+                otherwise: Joi.required(),
+            }),
+            dplDetectorId: Joi.number().when('runDetectors', {
+                is: Joi.exist(),
+                then: Joi.forbidden(),
+                otherwise: Joi.required(),
+            }),
+            dataPassId: Joi.number().allow(null),
+            simulationPassId: Joi.number().allow(null),
+            runDetectors: Joi.object().pattern(Joi.number(), Joi.array().items(Joi.number().required()).required()).optional(),
```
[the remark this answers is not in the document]
Jaden505: done

Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN apk add --no-cache \
     freetype=2.13.0-r5 \
     freetype-dev=2.13.0-r5 \
     harfbuzz=7.3.0-r0 \
-    ca-certificates=20241121-r0
```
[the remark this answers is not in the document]
Jaden505: done

lib/server/controllers/qcFlag.controller.js
```diff
@@ -147,49 +147,71 @@ const listSynchronousQcFlagsHandler = async (req, res) => {
...
+                    detectorIdentifier: { detectorId: flag.dplDetectorId },
+                    dataPassIdentifier: flag.dataPassId ? { id: flag.dataPassId } : null,
+                    simulationPassIdentifier: flag.simulationPassId ? { id: flag.simulationPassId } : null,
+                };
+
+                const [createdFlag] = await qcFlagService.create([flag], scope, relations);
+                createdFlags.push(createdFlag);
+            }
+
+            res.status(201).json({ data: createdFlags.length === 1 ? createdFlags[0] : createdFlags });
```
martinboulais, line 214: For the API it make sense to handle different approach as input (either directly runNumber/detectorId or a list of them) because it makes it more robust, and client will probably only use either one of them or the other, not both in the same code. 
But here, its make it more complex for clients to handle: independently of what they send, they will have to handle two different response type, all the time because they can't know in advance what you will return.
Here you should return always an array, and update the existing frontend (and tests) to handle it (simply checking that length is 1 and using first element for example).

test/api/qcFlags.test.js
```diff
@@ -326,6 +326,36 @@ module.exports = () => {
             }
         });
 
+        it('should successfully create multiple QC flag instances for data pass', async () => {
```
martinboulais, line 329: You also need to test that you can't specify mutually excluding parameters at the same time.

martinboulais (changes requested):

lib/server/controllers/qcFlag.controller.js
Jaden505: done

test/api/qcFlags.test.js
Jaden505: done

martinboulais (approved):

PR #1848, opened 2025-02-10
[O2B-1427] Store trigger configuration per runs

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- Display trigger configuration in a dedicated tab in run details page

- Added c++ API to update runs raw trigger configuration

Changes made to the database:
- Added column to store raw trigger configuration to runs table


cxx-client/include/BookkeepingApi/RunServiceClient.h
```diff
@@ -0,0 +1,29 @@
...
+#include <string>
+
+namespace o2::bkp::api
+{
+class RunServiceClient
+{
+ public:
+  virtual ~RunServiceClient() = default;
+
+  virtual void setRawTriggerConfiguration(int runNumber, std::string rawTriggerConfiguration) = 0;
```
graduta, line 25: I am not sure about the naming here. While `CTP` is indeed a trigger, we could also have `LTU` as a trigger. 
Is this configuration not specific to CTP only?

graduta (commented):
Except the name convention for the data that is to be stored, PR LGTM.

cxx-client/include/BookkeepingApi/RunServiceClient.h
martinboulais: Roman confirmed me that this is CTP only, I renamed accordingly. Thanks!

graduta (approved):

PR #1878, opened 2025-04-02
[O2B-1440] Use trigger to update flags on run boundaries change

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- NA

- NA

Changes made to the database:
- Added migration adding trigger which truncate QC flags boundaries when run timestamps are changed


martinboulais (changes requested):
Thanks for this PR, the changes looks good to me!
It might make sense to simply put a few tests (for example in run service, on run update?) to test these behaviors, what do you think?
It might also make sense to document this, either in bkp docs or in Jian's codimd.

martinboulais (approved):

PR #1886, opened 2025-04-10
[O2B-1437] Add basic authentication to gRPC

martinboulais:
#### I have a JIRA ticket
- [X] PR labels are selected

Notable changes for users:
- N/A

- Added possibility to provide authenticated gRPC API

Changes made to the database:
- N/A

### DIFF
--- a/cxx-client/src/BkpClientFactory.cxx
+++ b/cxx-client/src/BkpClientFactory.cxx
@@ -20,11 +20,24 @@
 using std::make_unique;
 using std::string;
 using std::unique_ptr;
+using grpc::ClientContext;
 
 namespace o2::bkp::api
 {
-unique_ptr<BkpClient> BkpClientFactory::create(const string& gRPCUri)
+unique_ptr<BkpClient> BkpClientFactory::create(const std::string& gRPCUri) {
+  auto clientContextFactory = []() {
+    return make_unique<ClientContext>();
+  };
+  return make_unique<grpc::GrpcBkpClient>(gRPCUri, clientContextFactory);
+}
+
+unique_ptr<BkpClient> BkpClientFactory::create(const string& gRPCUri, const string& token)
 {
-  return make_unique<grpc::GrpcBkpClient>(gRPCUri);
+  auto clientContextFactory = [token]() {
+    auto clientContext = make_unique<ClientContext>();
+    clientContext->AddMetadata("authorization", "Bearer " + token);
+    return clientContext;
+  };
+  return make_unique<grpc::GrpcBkpClient>(gRPCUri, clientContextFactory);
 }
 } // namespace o2::bkp::api
--- a/cxx-client/src/grpc/GrpcBkpClient.cxx
+++ b/cxx-client/src/grpc/GrpcBkpClient.cxx
@@ -20,6 +20,7 @@
 
 using grpc::Channel;
 
+using grpc::ClientContext;
 using grpc::CreateChannel;
 using grpc::InsecureChannelCredentials;
 using o2::bkp::api::FlpServiceClient;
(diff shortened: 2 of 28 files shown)


cxx-client/src/grpc/services/GrpcFlpServiceClient.cxx
```diff
@@ -48,7 +47,7 @@ void GrpcFlpServiceClient::updateReadoutCountersByFlpNameAndRunNumber(
   request.set_nrecordingbytes(nRecordingBytes);
   request.set_nfairmqbytes(nFairMQBytes);
 
-  auto status = mStub->UpdateCounters(&context, request, &updatedFlp);
+  auto status = mStub->UpdateCounters(mClientContextFactory().get(), request, &updatedFlp);
```
justonedev1, line 50: I precede this remark with a note that I am not exactly sure, how gRPC in c++ works as it was while since I use it.

Do I understand correctly, that you are creating `unique_ptr`, getting raw pointer from it and passing it somewhere else? Are you sure that this works without crashing? Because `unique_ptr` gets destroyed when the function call `UpdateCounters` ends and if the context is stored somewhere you might get double free (if the gRPC frees it automatically) or use after free if gRPC does not.

This comment is relevant for multiple places in this PR

cxx-client/src/grpc/GrpcBkpClient.cxx
```diff
@@ -37,14 +38,18 @@ using services::GrpcQcFlagServiceClient;
...
-  auto channel = CreateChannel(uri, InsecureChannelCredentials());
-  mFlpClient = make_unique<GrpcFlpServiceClient>(channel);
-  mDplProcessExecutionClient = make_unique<GrpcDplProcessExecutionClient>(channel);
-  mQcFlagClient = make_unique<GrpcQcFlagServiceClient>(channel);
-  mCtpTriggerCountersClient = make_unique<GrpcCtpTriggerCountersServiceClient>(channel);
-  mRunClient = make_unique<GrpcRunServiceClient>(channel);
+  auto channel = CreateChannel(
+    uri,
+    InsecureChannelCredentials()
+  );
```
justonedev1, line 46: Why the change of formatting here?

cxx-client/src/BkpClientFactory.cxx
```diff
@@ -20,11 +20,24 @@
...
+using grpc::ClientContext;
 
 namespace o2::bkp::api
 {
-unique_ptr<BkpClient> BkpClientFactory::create(const string& gRPCUri)
+unique_ptr<BkpClient> BkpClientFactory::create(const std::string& gRPCUri) {
+  auto clientContextFactory = []() {
+    return make_unique<ClientContext>();
+  };
+  return make_unique<grpc::GrpcBkpClient>(gRPCUri, clientContextFactory);
```
justonedev1, line 31: This comment is really just a remark and you don't have to address it:

I personally prefer to create the lambda directly as a function argument:
```
return make_unique<grpc::GrpcBkpClient>(gRPCUri, [](){return make_unique<ClientContext>();});
```
reason for this is I find it more readable as a one liner because it is not lvalue, which means that we are not expecting something else from this variable. Moreover the lambda is a rvalue here you don't even copy here anything etc... 

But it does not matter here, so your code is totally fine.

graduta (commented):
For JS part, lgtm

justonedev1 (commented):
I took a look and left one comment about part of code, that I either don't understand well enough or it is really wrong.

cxx-client/src/BkpClientFactory.cxx
martinboulais: I understand what you mean, thanks for the hint I changed it

cxx-client/src/grpc/services/GrpcFlpServiceClient.cxx
justonedev1: I'd prefer you change the code to explicitly creating context how I suggested earlier. No reasoning is necessary in code like that and function wise the code is identical.

PR #1931, opened 2025-07-16
[O2B-1452] Add filter option by creation date interval for retrieving environments

graduta:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- GET API endpoint `environments` now allows for retrieving them filtered based on creation date time interval [from, to]

-

Changes made to the database:
-

### DIFF
--- a/lib/usecases/environment/GetAllEnvironmentsUseCase.js
+++ b/lib/usecases/environment/GetAllEnvironmentsUseCase.js
@@ -85,9 +85,16 @@ class GetAllEnvironmentsUseCase {
                 currentStatus: currentStatusExpression,
                 statusHistory,
                 runNumbers: runNumbersExpression,
+                created,
             } = filter;
 
             const filterQueryBuilder = prepareQueryBuilder();
+
+            if (created) {
+                const from = created.from != null ? created.from : 0;
+                const to = created.to != null ? created.to : Date.now();
+                filterQueryBuilder.where('createdAt').between(from, to);
+            }
 
             if (idsExpression) {
                 const filters = idsExpression.split(',').map((id) => id.trim());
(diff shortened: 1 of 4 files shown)


lib/usecases/environment/GetAllEnvironmentsUseCase.js
```diff
@@ -85,10 +85,17 @@ class GetAllEnvironmentsUseCase {
                 currentStatus: currentStatusExpression,
                 statusHistory,
                 runNumbers: runNumbersExpression,
+                created,
             } = filter;
```
xsalonx, line 89: I would rename it, e.g. `createdFilter`, `createdExperssion`

lib/usecases/environment/GetAllEnvironmentsUseCase.js
```diff
@@ -85,10 +85,17 @@ class GetAllEnvironmentsUseCase {
...
 
             const filterQueryBuilder = prepareQueryBuilder();
 
+            if (created) {
+                const from = created.from != null ? created.from : 0;
+                const to = created.to != null ? created.to : Date.now();
```
xsalonx, line 95: I think it should be `undefined`, but  not `null`

lib/usecases/environment/GetAllEnvironmentsUseCase.js
graduta: For using 'expression' in the name, as far as I can see from the code, the convention is to use it when it is a parseable value. Such as a list of runNumbers or ids as a string separated by a comma. 
If we are looking at `GetAllRunsUseCase` or `GetAllLogsUseCase`, you will see the convention for naming filters from/to is the same as this one: `created`, `o2start`, `o2stop`, etc.

xsalonx: ok

lib/usecases/environment/GetAllEnvironmentsUseCase.js
graduta: Yes, codebase wise it is being use `!==` undefined but I thought it was an oversight due to the fact that this service can be called by another so in theory a null value could be passed. This check is for service check as `GetAllEnvironmentsUseCase` may be called by other services which did not validate the input.
     For example `GetLogTreeUseCase` and `GetLogsByTagUseCase` are both calling `GetAllLogsUseCase`. 

For consistency, I will move to using the convention

But for your second argument on Joi, the defined DTO schema will not allow for either `null` or `undefined`. But this check is only used in the controller. It will not prevent other services passing incorrect filter values to this service.

xsalonx (approved):

PR #1957, opened 2025-08-06
[O2B-1460] Filtering on Runs per LHC Period overview page

xsalonx:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Add filters to Runs Per LHC Period ovreview page: by run, fill, start, stop, L3/dipole, INEL values

- Refactor RunsPerLhcPeriodModel to use LHC Period ID instead of its name

Changes made to the database:
- NA

### DIFF
--- a/test/public/runs/runsPerLhcPeriod.overview.test.js
+++ b/test/public/runs/runsPerLhcPeriod.overview.test.js
@@ -193,6 +193,7 @@ module.exports = () => {
         // First export
         await pressElement(page, EXPORT_RUNS_TRIGGER_SELECTOR, true);
         await page.waitForSelector('select.form-control', { timeout: 200 });
+        await page.waitForSelector('option[value=runNumber]', { timeout: 200 });
         await page.select('select.form-control', 'runQuality', 'runNumber', 'definition', 'lhcPeriod');
         await expectInnerText(page, '#send:enabled', 'Export');
 


lib/public/views/Runs/RunPerPeriod/RunsPerLhcPeriodOverviewPage.js
```diff
@@ -54,42 +60,27 @@ export const RunsPerLhcPeriodOverviewPage = ({ runs: { perLhcPeriodOverviewModel
     const {
         items: remoteRuns,
         detectors: remoteDetectors,
-        lhcPeriodName,
+        lhcPeriod: remoteLhcPeriod,
```
graduta, line 63: These are actually the statistics and not the lhcPeriod, right?
Could you rename accordingly here and at L100, L104? 
This in turn will also help with  L116 where you reference `lhcPeriod.lhcPeriod.name`

graduta (changes requested):
Given that the `runs` request already contains both the `lhcPeriodId` and `lhcPeriodName` and on top of that the table bug in which the request for `runs` and now for `lhcPeriodStatistics` are duplicated, is it worth doing a statistics request here on load?
I cannot see any other data being used from this request that would be displayed on the page

lib/public/views/Runs/RunPerPeriod/RunsPerLhcPeriodOverviewPage.js
xsalonx: applied

graduta (dismissed):

graduta (approved):

graduta (dismissed):

graduta (approved):

PR #1964, opened 2025-08-19
[O2B-1470] Overflow balloon content not clickable

xsalonx:
#### I DON'T have JIRA ticket
- [x] tests are added

Notable changes for users:
-  Whole content of overflow balloon is now clickable

- Refactor overflowBalloon component

Changes made to the database:
- NA

### DIFF
--- a/lib/public/views/Runs/RunPerPeriod/RunsPerLhcPeriodOverviewModel.js
+++ b/lib/public/views/Runs/RunPerPeriod/RunsPerLhcPeriodOverviewModel.js
@@ -48,8 +48,38 @@ export class RunsPerLhcPeriodOverviewModel extends RunsWithQcModel {
 
         this._syncDetectors$.bubbleTo(this);
 
+        this._lhcPeriodId = null;
+        this._lhcPeriodStatistics$ = new ObservableData(RemoteData.notAsked());
+        this._lhcPeriodStatistics$.bubbleTo(this);
+
         this._tabbedPanelModel = new RunsPerLhcPeriodTabbedPanelModel();
         this._tabbedPanelModel.bubbleTo(this);
+    }
+
+    /**
+     * Fetch LHC period data which runs are fetched
+     * @return {Promise<void>} promise
+     */
+    async _fetchLhcPeriod() {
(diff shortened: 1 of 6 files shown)


lib/public/components/common/popover/overflowBalloon.js
```diff
@@ -53,16 +53,33 @@ export const overflowBalloon = (content, options = null) => {
              */
             onTriggerNodeChange: function (previousTriggerNode, newTriggerNode) {
                 if (previousTriggerNode) {
-                    const node = stretch ? previousTriggerNode.parentNode : previousTriggerNode;
-
-                    node.removeEventListener('mouseover', this.showPopover);
-                    node.removeEventListener('mouseleave', this.hidePopover);
+                    previousTriggerNode.removeEventListener('mouseover', this.prevousNodeMouseoverCallback);
```
graduta, line 56: Typo on `this.prevousNodeMouseoverCallback`

lib/public/components/common/popover/overflowBalloon.js
```diff
@@ -53,16 +53,33 @@ export const overflowBalloon = (content, options = null) => {
...
                 if (newTriggerNode) {
                     const node = stretch ? newTriggerNode.parentNode : newTriggerNode;
+                    const key = newTriggerNode.getAttribute('data-popover-key');
+
+                    const newNodeMouseoverCallback = () => {
+                        this.showPopover();
+                        const selector = `.popover[data-popover-key="${key}"]`;
+                        const contentNode = document.querySelector(selector);
+
+                        contentNode.addEventListener('mouseover', () => {
```
graduta, line 69: Can it be that the `querySelector` will fail because of invalid id? 
Should it be checked that contentNode is not null before adding listeners?

graduta, line 69: Looking at the JS MDN docs, looks like the method will add to a list of eventlisteners and not replace them by name.
Should it not be better to first attempt to remove existing listeners and then add them? Currently when I logged locally, the listeners are chained and called tens of times because of this

graduta (changes requested):

graduta (approved):

PR #2083, opened 2026-02-20
[O2B-1534] Migrate log overview to use filtering model pattern

NarrowsProjects:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Individual from-to selectors on the logs page have been replaced with a single field that will open the standard time range selector
- All open-input forms apply their filters when focus is lost on them.

- All filters have been moved to a `filteringModel` in `logsOverviewModel`
- The filters for runs, lhcfills and environments have been renamed under the hood to match getAllRunsUseCase:
  - runs =>                 runNumbers
  - lhcFills =>             fillNumbers
  - environments => environmentIds

Changes made to the database:
- none


lib/public/components/Filters/LogsFilter/author/authorFilter.js
```diff
@@ -55,11 +43,16 @@ export const excludeAnonymousLogAuthorToggle = (authorFilterModel) => switchInpu
 /**
  * Returns a authorFilter component with text input, reset button, and anonymous exclusion button.
  *
- * @param {LogModel} logModel the log model object
- * @returns {Component} the author filter component
+ * @param {LogsOverviewModel} logsOverviewModel the log overview model
```
isaachilly: The overviewModel param doesn't exist in the func.

lib/public/components/Filters/LogsFilter/author/authorFilter.js
```diff
@@ -55,11 +43,16 @@ export const excludeAnonymousLogAuthorToggle = (authorFilterModel) => switchInpu
...
+ * @param {FilteringModel} logsOverviewModel.filteringModel the runs overview model
+ * @return {Component} the filter component
  */
-export const authorFilter = ({ authorFilter }) => h('.flex-row.items-center.g3', [
-    authorFilterTextInput(authorFilter),
-    resetAuthorFilterButton(authorFilter),
-    excludeAnonymousLogAuthorToggle(authorFilter),
+export const authorFilter = ({ filteringModel }) => h('.flex-row.items-center.g3', [
```
isaachilly: It would be more consistent with other `ActiveColumns` Filter setups to get the matching filter's model in `ActiveColumns` and pass only that to the Filter. And we sent the function only the specific data it needs.

lib/public/components/Filters/LogsFilter/author/authorFilter.js
```diff
@@ -36,7 +24,7 @@ const authorFilterTextInput = (authorFilterModel) => h('input.w-40', {
  */
 const resetAuthorFilterButton = (authorFilterModel) => h(
     '.btn.btn-pill.f7',
-    { disabled: authorFilterModel.isEmpty, onclick: () => authorFilterModel.clear() },
+    { disabled: authorFilterModel.isEmpty, onclick: () => authorFilterModel.reset() },
```
isaachilly: `.clear()` reloaded the table's items because it notified. This is not the case _currently_ with `.reset()` of `RawTextFilterModel` thus, when you click the cross button, the author text is emptied, but the table remains the same, it is not refreshed as it should be.

lib/public/components/Filters/LogsFilter/author/authorFilter.js
NarrowsProjects: So, with js javadoc conventions, documenting logsOverviewModel is correct, but I realize now that we don't (generally) do this in the existing codebase, so:
Taken care of

lib/public/components/Filters/LogsFilter/author/authorFilter.js
NarrowsProjects: Taken care of

lib/public/components/Filters/LogsFilter/author/authorFilter.js
NarrowsProjects: Taken care off

lib/public/components/Filters/common/filters/RawTextFilterModel.js
```diff
@@ -17,8 +17,12 @@ export class RawTextFilterModel extends FilterModel {
     /**
      * @inheritDoc
      */
-    reset() {
+    reset(notify = false) {
```
isaachilly: TBD offline with result commented here.

lib/public/components/Filters/LogsFilter/author/authorFilter.js
```diff
@@ -14,31 +14,16 @@
...
  * @param {AuthorFilterModel} authorFilterModel The author filter model object
  * @return {Component} A button that can be used to reset the author filter
  */
-const resetAuthorFilterButton = (authorFilterModel) => h(
-    '.btn.btn-pill.f7',
-    { disabled: authorFilterModel.isEmpty, onclick: () => authorFilterModel.clear() },
-    iconX(),
-);
+const resetAuthorFilterButton = (authorFilterModel) =>
+    h('.btn.btn-pill.f7', { disabled: authorFilterModel.isEmpty, onclick: () => authorFilterModel.reset(true) }, iconX());
```
isaachilly: True is still being sent to the function here, we are now overwriting it so don't need the param.

NarrowsProjects: Thanks for the catch

lib/public/components/Filters/LogsFilter/author/AuthorFilterModel.js
```diff
@@ -42,27 +42,20 @@ export class AuthorFilterModel extends FilterInputModel {
...
-     *
-     * @return {void}
+     * @inheritDoc
      */
-    clear() {
-        if (this.isEmpty) {
-            return;
-        }
-
+    reset() {
```
isaachilly: I think the clear method just needs reinstating because it handled the if empty and the notify. This will create a duplicate fetch because filteringModel will call reset on the model and it'll notify.

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
```diff
@@ -71,13 +68,15 @@ export const logsActiveColumns = {
...
-        filter: ({ titleFilter }) => textFilter(
-            titleFilter,
-            {
-                id: 'titleFilterText',
-                class: 'w-75 mt1',
-            },
-        ),
+
+        /**
+         * Title filter component
```
isaachilly, line 73: Title, contend and author do not have placeholder text and I think it would be good and consistent for them to have.

NarrowsProjects: Taken care of

lib/usecases/log/GetAllLogsUseCase.js
```diff
@@ -111,74 +123,47 @@ const applyFilter = async (dataSource, queryBuilder, filter) => {
         queryBuilder.where('id').oneOf(...logIds);
     }
 
-    if (filter.run?.values?.length > 0) {
+    if (runNumbers?.length > 0) {
```
isaachilly: Can this just be if runNumbers? Check the other filter ifs.

test/public/logs/overview.test.js
```diff
@@ -154,17 +154,25 @@ module.exports = () => {
...
-        await waitForTableLength(page, 1);
+        
+        await fillInput(page, fromDateSelector, limit, ['change']);
+        await fillInput(page, toDateSelector, limit, ['change']);
+        await fillInput(page, fromTimeSelector, '11:00', ['change']);
+        await fillInput(page, toTimeSelector, '12:00', ['change']);
 
-        await pressElement(page, '#reset-filters');
+        await waitForTableLength(page, 1);
+        await openFilteringPanel(page);
```
isaachilly: Is this line needed?

test/public/logs/overview.test.js
```diff
@@ -154,17 +154,25 @@ module.exports = () => {
...
+        await resetFilters(page);
```
isaachilly: Could we make use of before/after each functions to handle resetting filters etc?

lib/usecases/log/GetAllLogsUseCase.js
NarrowsProjects: I suppose you are right, I just didn't question the original logic

test/public/logs/overview.test.js
```diff
@@ -91,57 +95,56 @@ module.exports = () => {
...
-        await pressElement(page, '#reset-filters');
+        await resetFilters(page);
     });
 
     it('should successfully provide an easy-to-access button to filter in/out anonymous logs', async () => {
         // Close the filter panel
         await pressElement(page, '#openFilterToggle');
         await waitForTableTotalRowsCountToEqual(page, 119);
-
+        // await new Promise((res, _rej) => setTimeout(() => res(), 2000))
```
isaachilly: I'm assuming leftover things here.

test/public/logs/overview.test.js
```diff
@@ -153,376 +156,380 @@ module.exports = () => {
...
-
-    it('should successfult navigate to run details via popover', async () => {
-        await goToPage(page, 'log-overview');
-        const cellSelector = 'td#row119-runs'
-        await page.hover(cellSelector);
-        const popoverSelector = await getPopoverSelector(await page.waitForSelector(`${cellSelector} .popover-trigger`));
-        await waitForNavigation(page, () => pressElement(page, `${popoverSelector} a`))
-        expectUrlParams(page, { page: 'run-detail', runNumber: 2 })
-    });
+    // it('can filter by creation date', async () => {
```
isaachilly: An accidental commit of commented out tests.

NarrowsProjects: lovely
Also, taken care of

isaachilly (changes requested):

lib/public/components/Filters/LogsFilter/author/AuthorFilterModel.js
NarrowsProjects: Taken care of

test/public/logs/overview.test.js
NarrowsProjects: Yes, but the line shouldn't be there, the next test needs the filteringpanel to be open

test/public/logs/overview.test.js
NarrowsProjects: Taken care of

lib/public/components/Filters/common/filters/RawTextFilterModel.js
NarrowsProjects: Outdated

test/public/logs/overview.test.js
NarrowsProjects: Taken care of

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
```diff
@@ -71,13 +68,18 @@ export const logsActiveColumns = {
...
+         *
+         * @param {FilteringModel} logOverviewModel.filteringModel filtering model
+         * @return {Component} the filter component
+         */
+        filter: ({ filteringModel }) =>
+            rawTextFilter(
+                filteringModel.get('title'),
+                { id: 'titleFilterText', classes: ['w-75 mt1'], placeholder: 'e.g. Report on runs: ...' },
```
isaachilly: I think the widths of the filters inputs could be adjusted to match other pages with filters, as in full width.

test/api/logs.test.js
```diff
@@ -233,7 +233,7 @@ module.exports = () => {
         });
```
isaachilly, line 234: Sorting by runs nor environments is tested as well as filtering by rootOnly. Would be good I think to take this chance to add these tests now.

isaachilly (changes requested):

lib/public/components/Filters/LogsFilter/author/authorFilter.js
```diff
@@ -55,11 +40,11 @@ export const excludeAnonymousLogAuthorToggle = (authorFilterModel) => switchInpu
...
- * @returns {Component} the author filter component
+ * @param {AuthorFilterModel} authorFilterModel the authorFilterModel
+ * @return {Component} the filter component
  */
-export const authorFilter = ({ authorFilter }) => h('.flex-row.items-center.g3', [
-    authorFilterTextInput(authorFilter),
-    resetAuthorFilterButton(authorFilter),
-    excludeAnonymousLogAuthorToggle(authorFilter),
+export const authorFilter = (authorFilterModel) => h('.flex-row.items-center.g3', [
+    rawTextFilter(authorFilterModel, { classes: ['w-60'], id: 'authorFilterText', value: authorFilterModel.raw, placeholder: 'e.g. John Doe' }),
```
isaachilly: On my computer w-60 makes the exclude Anonymous wrap, w-50 could be better if this is also the case on yours.

lib/usecases/log/GetAllLogsUseCase.js
isaachilly: Yes, that's fine. I'm trying to strike a balance of making suggestions directly related to the ticket versus ones that are related but not absolutely needing to be addressed by the ticket.

lib/public/components/Filters/LogsFilter/author/authorFilter.js
NarrowsProjects: 60 works fine on my screen size, but It is good to err on the side of causion.
Taken care of

test/api/logs.test.js
NarrowsProjects: Taken care off

isaachilly: I think checking every item returned is a root log is enough rather than having to deal with counts?

If you indeed want to check that there are even logs with children to be filtered out, you could do a filter for children (rootLogId !== id) and check this is > 0.

NarrowsProjects: Fair

isaachilly: Do you need the limit now and in turn the comment explaining the limit?

NarrowsProjects: I suppose not

isaachilly (dismissed):

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
```diff
@@ -150,7 +172,21 @@ export const logsActiveColumns = {
...
         format: formatRunsList,
-        filter: runsFilter,
+
+        /**
+         * Runs filter component
+         *
+         * @param {FilteringModel} logOverviewModel.filteringModel filtering model
+         * @return {Component} the filter component
+         */
+        filter: ({ filteringModel }) => rawTextFilter(
```
graduta: What about re-using the `runNumbersFilter` in this case?

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
```diff
@@ -167,7 +203,21 @@ export const logsActiveColumns = {
...
-        filter: environmentFilter,
+
+        /**
+         * Environment filter component
+         *
+         * @param {FilteringModel} logOverviewModel.filteringModel filtering model
+         * @return {Component} the filter component
+         */
+        filter: ({ filteringModel }) => rawTextFilter(
+            filteringModel.get('environmentIds'),
```
graduta: Similar question, I see that runsActiveColumns also has the environmentIds as rawtextFilter
Could we perhaps extract as a reusable component similar to runNumbersFilter?

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
```diff
@@ -177,7 +227,21 @@ export const logsActiveColumns = {
...
         format: formatLhcFillsList,
-        filter: lhcFillsFilter,
+
+        /**
+         * LhcFills filter component
+         *
+         * @param {FilteringModel} logOverviewModel.filteringModel filtering model
+         * @return {Component} the filter component
+         */
+        filter: ({ filteringModel }) => rawTextFilter(
```
graduta: Same here

lib/usecases/log/GetAllLogsUseCase.js
```diff
@@ -111,74 +123,47 @@ const applyFilter = async (dataSource, queryBuilder, filter) => {
         queryBuilder.where('id').oneOf(...logIds);
     }
 
-    if (filter.run?.values?.length > 0) {
+    if (runNumbers) {
```
graduta, line 126: Should we not still check if array is not empty?

lib/usecases/log/GetAllLogsUseCase.js
```diff
@@ -111,74 +123,47 @@ const applyFilter = async (dataSource, queryBuilder, filter) => {
...
-        }
+        let logRuns = await LogRunsRepository.findAllAndGroup(runQueryBuilder);
+        logRuns = logRuns.filter((logRun) => runNumbers.every((runNumber) => logRun.runNumbers.includes(runNumber)));
 
         const logIds = logRuns.map((logRun) => logRun.logId);
         queryBuilder.where('id').oneOf(...logIds);
     }
 
-    if (filter.lhcFills?.values?.length > 0) {
+    if (fillNumbers) {
```
graduta, line 140: Should we not still check if array is not empty?
We could define on L54 as `fillNumbers = []` and straight away check length perhaps?

lib/usecases/log/GetAllLogsUseCase.js
```diff
@@ -111,74 +123,47 @@ const applyFilter = async (dataSource, queryBuilder, filter) => {
...
+            fillNumbers.every((fillNumber) => logLhcFill.fillNumbers.includes(fillNumber)));
 
         const logIds = logLhcFills.map((logLhcFill) => logLhcFill.logId);
 
         queryBuilder.where('id').oneOf(...logIds);
     }
 
-    if (filter.environments?.values?.length > 0) {
-        const validEnvironments = await EnvironmentRepository.findAll({ where: { id: { [Op.in]: filter.environments.values } } });
+    if (environmentIds) {
```
graduta, line 156: Same here

lib/domain/dtos/GetAllLogsDto.js
```diff
@@ -17,32 +17,25 @@ const PaginationDto = require('./PaginationDto');
...
-const RunFilterDto = Joi.object({
-    values: CustomJoi.stringArray().items(EntityIdDto).single().required(),
-    operation: Joi.string().valid('and', 'or').required(),
-});
-
-const LhcFillFilterDto = Joi.object({
-    values: CustomJoi.stringArray().items(EntityIdDto).single().required(),
-    operation: Joi.string().valid('and', 'or').required(),
-});
+const RunFilterDto = CustomJoi.stringArray().items(EntityIdDto).single();
```
graduta, line 21: I see these `RunFilter`, Environment, LHC being used across DTOs with different matching criterias.
What about extracting this in its own DTO and reuse across DTOs?

graduta (changes requested):

lib/usecases/log/GetAllLogsUseCase.js
NarrowsProjects: From what I can remember, there is no way for runNumber to arrive here as an empty array.
The Joi DTO requires it to be a string of numbers, and an empty string will be rejected

lib/usecases/log/GetAllLogsUseCase.js
NarrowsProjects: As above

lib/usecases/log/GetAllLogsUseCase.js
NarrowsProjects: As above

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
NarrowsProjects: I don't see much use in making or using individual components for each filter that uses rawTextFilter that are virtually Identical.

What I do think is a good idea now that you mention it is to make a new component that would cover most of these usecases and change it for special cases such as authorFilter. Something like

```js
export const textInputFilter =
    (filteringModel, key, placeholder) => rawTextFilter(filteringModel.get(key), { classes: ['w-100', `key-${filter}`], placeholder });
```

Which would then shorten these filter creations to:
```js
filter: ({ filteringModel }) => textInputFilter(filteringModel, 'runNumbers', 'e.g. 553203, 553221, ...'),
```

lib/domain/dtos/GetAllLogsDto.js
NarrowsProjects: I will have a look to see if they can be standardised

lib/usecases/log/GetAllLogsUseCase.js
NarrowsProjects: I've done some manual tests, and I can't find a way to reach that line as an empty array

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
NarrowsProjects: As above

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
NarrowsProjects: As above

lib/usecases/log/GetAllLogsUseCase.js
graduta: @NarrowsProjects what about crafting a CURL/POSTMAN request to simulate the users using scripts for fetching data via the HTTP API rather than front-end?

lib/domain/dtos/GetAllLogsDto.js
NarrowsProjects: So, I have attempted to standardize one of the Joy schemas, and It is definitely a lot of work. 

I would argue that it should be added as a task to this ticket:
[O2B-1579](https://its.cern.ch/jira/browse/O2B-1579)

Instead of in this PR.

lib/usecases/log/GetAllLogsUseCase.js
NarrowsProjects: That is what I did, from the best of my knowelege it will always be caught by the Joi schema validation

graduta: You are correct. My concern was with regards to the services that are being used by the gRPC based controllers. Those are not using the DTO validators but with the classes named `Get...UseCase`, they expect a DTO as parameter so your evaluation is correct

lib/public/views/LhcFills/ActiveColumns/lhcFillsActiveColumns.js
```diff
@@ -24,11 +24,10 @@ import { infologgerLinksComponents } from '../../../components/common/externalLi
 import { formatBeamType } from '../../../utilities/formatting/formatBeamType.js';
 import { frontLink } from '../../../components/common/navigation/frontLink.js';
 import { toggleStableBeamOnlyFilter } from '../../../components/Filters/LhcFillsFilter/stableBeamFilter.js';
-import { fillNumberFilter } from '../../../components/Filters/LhcFillsFilter/fillNumberFilter.js';
 import { durationFilter } from '../../../components/Filters/LhcFillsFilter/durationFilter.js';
 import { beamTypeFilter } from '../../../components/Filters/LhcFillsFilter/beamTypeFilter.js';
-import { schemeNameFilter } from '../../../components/Filters/LhcFillsFilter/schemeNameFilter.js';
```
graduta, line 30: Should this file not be removed the same as `runNumbersFilter`? I do not see it being used anywhere else

lib/public/views/LhcFills/ActiveColumns/lhcFillsActiveColumns.js
```diff
@@ -24,11 +24,10 @@ import { infologgerLinksComponents } from '../../../components/common/externalLi
 import { formatBeamType } from '../../../utilities/formatting/formatBeamType.js';
 import { frontLink } from '../../../components/common/navigation/frontLink.js';
 import { toggleStableBeamOnlyFilter } from '../../../components/Filters/LhcFillsFilter/stableBeamFilter.js';
-import { fillNumberFilter } from '../../../components/Filters/LhcFillsFilter/fillNumberFilter.js';
```
graduta, line 27: Should this file not be removed?

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
```diff
@@ -15,19 +15,16 @@ import { h } from '/js/src/index.js';
 import { iconCommentSquare, iconPaperclip } from '/js/src/icons.js';
 
 import { authorFilter } from '../../../components/Filters/LogsFilter/author/authorFilter.js';
-import createdFilter from '../../../components/Filters/LogsFilter/created.js';
```
graduta, line 18: Should be removed?

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
```diff
@@ -15,19 +15,16 @@ import { h } from '/js/src/index.js';
...
-import runsFilter from '../../../components/Filters/LogsFilter/runs.js';
```
graduta, line 19: Should be removed?

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
```diff
@@ -15,19 +15,16 @@ import { h } from '/js/src/index.js';
...
 import { formatTimestamp } from '../../../utilities/formatting/formatTimestamp.js';
 import { frontLink } from '../../../components/common/navigation/frontLink.js';
 import { frontLinks } from '../../../components/common/navigation/frontLinks.js';
 import { tagFilter } from '../../../components/Filters/common/filters/tagFilter.js';
 import { formatRunsList } from '../../Runs/format/formatRunsList.js';
 import { profiles } from '../../../components/common/table/profiles.js';
-import { textFilter } from '../../../components/Filters/common/filters/textFilter.js';
-import { environmentFilter } from '../../../components/Filters/LogsFilter/environments.js';
```
graduta, line 27: ^

lib/public/views/Logs/ActiveColumns/logsActiveColumns.js
```diff
@@ -15,19 +15,16 @@ import { h } from '/js/src/index.js';
...
 import { formatLhcFillsList } from '../../LhcFills/format/formatLhcFillsList.js';
-import { lhcFillsFilter } from '../../../components/Filters/LogsFilter/lhcFill.js';
```
graduta, line 29: ^

graduta (changes requested):

graduta (dismissed):

graduta (approved):

isaachilly (approved):

PR #2106, opened 2026-03-10
[O2B-1552] Move ad hoc filters inside runs overviews model to filtering model

NarrowsProjects:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- The monte carlo reproducable as not bad filters from the runs per .... overview pages has been changed to also be cleared when the 'reset all filters' button is pressed to be in line with all other filters.

**Filter restructuring**

- `ddflpFilter`, `dcsFilter`, and `epnFilter` have been moved to the `filteringModel`.
- `detectorsQc[_id][notBadFraction]` filters have been moved to a new model: `MultiCompositeFilterModel`.
- `gaq[notBadFraction]` now uses its own filter model with unique normalization logic: `GaqFilterModel`.

**New filter models**
- **`GaqFilterModel`**
  - Dedicated model for `gaq[notBadFraction]`.
  - Implements custom normalization logic.

- **`ToggleFilterModel`**
  - Introduced for toggle-based filters.
  - `mcReproducibleAsNotBad` has been converted to use this model.
- **`MultiCompositeFilterModel`**
  - New flexible model:
    - Can be instantiated with filters.
    - Supports adding filters dynamically after instantiation.
  - Computes normalized values based on:
    - Insertion keys
    - Normalized child filters

**Shared filter behavior**
- `mcReproducibleAsNotBad`:
  - Converted into a `ToggleFilterModel`.
  - Integrated into both:
    - `GaqFilterModel`
    - `MultiCompositeFilterModel`
  - **Important:** both models must use the same instance for correct behavior.

**New components**
- `RadioButtonFilter`
  - Replaces filters associated with the restructured filter models.
- `ToggleFilter` (component + model)
  - Introduced alongside `ToggleFilterModel`.

**Renaming**

- `detectorsQc[_id][notBadFraction]` → `detectorsQcNotBadFraction[_id]`

  - Reduces nesting within the `filterModel`.

**Implementation note**

- This approach avoids:

  - Creating a custom wrapper around `NumericalComparisonFilterModel` solely for normalization, or
  - Nesting `MultiCompositeFilterModel` within another `MultiCompositeFilterModel` to achieve automatic normalization.

Changes made to the database:
- N/A


isaachilly (changes requested):

test/public/runs/overview.test.js
```diff
@@ -600,7 +600,7 @@ module.exports = () => {
 
         it('Should successfully filter runs by their trigger value', async () => {
             await navigateToRunsOverview(page);
-            const filterInputSelectorPrefix = '#triggerValueCheckbox';
+            const filterInputSelectorPrefix = '#checkboxes-checkbox-';
```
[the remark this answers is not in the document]
NarrowsProjects: Taken care of

lib/public/views/Runs/ActiveColumns/runsActiveColumns.js
```diff
@@ -534,14 +531,21 @@ export const runsActiveColumns = {
...
-        filter: triggerValueFilter,
+
+        /**
+         * TriggerValue filter component
+         *
+         * @param {RunsOverviewModel} runsOverviewModel the runs overview model
+         * @return {Component} the trigger value filter component
+         */
+        filter: ({ filteringModel }) => checkboxes(filteringModel.get('triggerValues')),
         format: (trgValue) => trgValue ? trgValue : '-',
```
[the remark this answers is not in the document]
NarrowsProjects: Taken care of

lib/public/components/Filters/common/filters/ToggleFilterModel.js
```diff
@@ -0,0 +1,63 @@
...
+
+        this._falseIsEmpty = falseIsEmpty;
+    }
+
+    /**
+     * Returns true if the current value is set to true
+     *
+     * @return {boolean} true if filter is stable beams only
+     */
+    isToggled() {
```
isaachilly: Could this be a getter? `isToggled` represents a state rather than an action, so I think it would be more natural, what do you think?

NarrowsProjects: It makes sense, especially since there is already precedence for this with how FilterModels have `isEmpty` as a getter

lib/public/components/Filters/common/filters/radioButtonFilter.js
```diff
@@ -0,0 +1,38 @@
...
+
+import { radioButton } from '../../../common/form/inputs/radioButton.js';
+import { h } from '/js/src/index.js';
+
+/**
+ * Radiobutton filter component
+ *
+ * @param {RadioSelectionModel} selectionModel the a selectionmodel
+ * @param {string} filterName the name of the filter
+ * @return {vnode} A number of radiobuttons corresponding with the selection options
```
isaachilly: `radiobuttons` think you forgot a space.

NarrowsProjects: Good catch

lib/domain/dtos/filters/RunFilterDto.js
isaachilly: Could you rename this to match the new name of the filter?

isaachilly: I have noticed that this dataPassIds change does not actually work as it is testing relatively here and not checking the parent, as demonstrated by the detectorsQcNotBadFraction using the helpers. 

It would be good to rectify this and also add a test in the api.

lib/public/components/Filters/common/filters/toggleFilter.js
```diff
@@ -0,0 +1,45 @@
...
+ * Display a toggle switch or radio buttons for toggle filters
+ *
+ * @param {ToggleFilterModel} toggleFilterModel a ToggleFilterModel
+ * @param {name} toggleFilterModel the name used to identify and label the filter
+ * @param {boolean} radioButtonMode define whether or not to return radio buttons or a switch.
+ * @returns {Component} the toggle switch
+ */
+export const toggleFilter = (toggleFilterModel, name, id, radioButtonMode = false) => {
+    if (radioButtonMode) {
+        return h('.form-group-header.flex-row.w-100', [
```
isaachilly: Given we are creating `toggleFilter`, which is definitely a nice abstraction, I looked at the classes here and `form-group-header` I don't think should be used. Neither in `toggleFilter.js` or in `radioButtonFilter.js`. From what I can tell it was put into `combinationOperatorChoiceComponent.js` to add some spacing between its radio buttons and then (in all current cases) a subsequent selection dropdown. But in the `toggleFilter.js` and `radioButtonFilter.js` these do not have any following elements and this is then excess margin.

lib/public/components/Filters/common/filters/ToggleFilterModel.js
```diff
@@ -0,0 +1,63 @@
...
+     * @return {void}
+     */
+    toggle() {
+        this.select({ value: !this.current });
+    }
+
+    /**
+     * @inheritdoc
+     */
+    get isEmpty() {
```
isaachilly, line 60: I think this could do with at least an explanatory comment as it is quite confusing with the false and true wording of everything.

lib/domain/dtos/filters/RunFilterDto.js
```diff
@@ -88,26 +88,28 @@ exports.RunFilterDto = Joi.object({
...
-            'dataPassIds',
-            {
-                is: Joi.array().length(1),
-                then: FloatComparisonDto,
-                otherwise: Joi.forbidden().error(new Error('Filtering by GAQ is enabled only when filtering with one dataPassId')),
-            },
-        ),
+        notBadFraction: FloatComparisonDto.custom((value, helpers) => {
+            // eslint-disable-next-line prefer-destructuring
+            const { dataPassIds } = helpers.state.ancestors[1];
```
isaachilly: So we don't start a habit (not that you have one at all) I'd just destructure here.

NarrowsProjects: I understand, but the way destructuring would be done here is like this:

```js
            const [, { dataPassIds }] = helpers.state.ancestors;
```

And I can see this being easily missable by people trying to read the code. 
However, we do already destructure like this in other files in the code, and disabling a eslint checking destructuring is done nowhere else.

So, I would like your thoughts on this.

NarrowsProjects: After thinking about it for a bit, I think your were right.

lib/public/components/Filters/common/filters/ToggleFilterModel.js
NarrowsProjects: Taken care of

isaachilly: Could this be made clearer with a different choice of words? I think people could get confused with all the falses and trues.

NarrowsProjects: I have now renamed the parameter to untoggledIsEmpty with the adjustments to the isEmpty getter and it's jsdoc

isaachilly: @graduta Me and Guust aggreed it would be good for you to give your opinion on this specifically. I am not convinced a property like `_emptyValue` set to True/False from constructor and compare to this in the `isEmpty()` function is not the more clear solution.

graduta: Hi, to me `isEmpty` refers and extends the selection model. Thus:
- false - not selected and so empty
- true - selected and thus not empty as it contain a value of `1/true`

graduta (approved):

PR #2142, opened 2026-04-21
[O2B-1556] Configure filtering model observability to set active filters to the url

NarrowsProjects:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- When configuring a filter a string version will then appear in the URL

- FilteringModel now has a router argument for its constructor.
- All pages that use a filteringModel must explicitly set filteringModel.pageIdentifier. This identifier determines if this specific filteringModel instance should sync its active filters to the URL, and which page’s URL should be used to initialize the filter values for that filteringModel.
  - This is currently done by passing the identifier page identifier via the constructor
- All overview page reset() functions have gained a clearUrl boolean argument, that determines if running the reset function should also remove the filters from the url.

Changes made to the database:
- N/A


test/public/defaults.js
```diff
@@ -667,14 +667,24 @@ module.exports.checkColumnBalloon = async (page, rowIndex, columnIndex) => {
  * @return {Promise} resolves once the value has been typed
  */
 module.exports.fillInput = async (page, inputSelector, value, events = ['input']) => {
-    await page.waitForSelector(inputSelector);
-    await page.evaluate((inputSelector, value, events) => {
+    await page.waitForFunction((inputSelector, value, events) => {
```
NarrowsProjects, line 670: To the reviewers:
  This is a stealth fix for a problem that happens for a couple of tests occasionally:

```js
     TypeError: Cannot set properties of null (setting 'value')
      at evaluate (evaluate at module.exports.fillInput (test/public/defaults.js:671:16), <anonymous>:2:22)
```

This happens because between waitForSelector and the evaluate the selector ceases to be valid.
The waitForFunction guard is more more robust waitForSelector in this scenario.

As this PR adds tests that make back-to-back use of `fillInput` I thought it appropriate to add it here.

graduta (approved):
I think it would be good if we also move away from page identifiers as hard-coded strings and use a map/enum for this. Could you please create a ticket so that we do not forget about this?

PR #2143, opened 2026-04-23
[O2B-1555] Add normalize setter functions to all classes that extend filter model and selection model

NarrowsProjects:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- Setting filters will now display them in the url

- All filters now have a normalized setter that will set the values of the filters based on the output of the normalize getter in query parameter form.

Changes made to the database:
- N/A


lib/public/components/common/selection/SelectionModel.js
```diff
@@ -332,6 +351,23 @@ export class SelectionModel extends Observable {
...
     }
 
+    /**
+     * Sets selected options based on a comma-seperated string.
+     * Accounts for the options being either RemoteData or an array.
+     *
+     * @return {string}
+     */
+    set normalized(value) {
+        const options = value.split(',').map((option) => ({ value: option }));
```
isaachilly: Could trimming the split values be good to catch user errors, or will it interfere with filters that should keep the whitespace?

lib/public/components/common/selection/SelectionModel.js
```diff
@@ -332,6 +351,23 @@ export class SelectionModel extends Observable {
         return this._defaultSelection;
     }
 
+    /**
+     * Sets selected options based on a comma-seperated string.
+     * Accounts for the options being either RemoteData or an array.
+     *
+     * @return {string}
+     */
+    set normalized(value) {
```
isaachilly, line 359: What happens if value is null or undefined? Could value ever be null or undefined? Maybe not in UI but maybe if passed in URL? What do you think? Goes for all the setters.

lib/public/components/common/selection/SelectionModel.js
```diff
@@ -243,14 +249,20 @@ export class SelectionModel extends Observable {
...
+     * Defines the list of available options and if there is a selection backlog, these will be applied
      *
      * @param {RemoteData<SelectionOption[], *>|SelectionOption[]} availableOptions the new available options
      * @return {void}
      */
     setAvailableOptions(availableOptions) {
         this._availableOptions = availableOptions;
         this.visualChange$.notify();
+
+        if (this._selectionBacklog.length) {
```
isaachilly, line 261: I think this leaves an edge case of attempting to apply the backlog before availableOptions has finished loading which will result in nothing being selected and then the backlog getting cleared.

lib/public/components/Filters/RunsFilter/MagnetsFilteringModel.js
```diff
@@ -11,86 +11,57 @@
...
      */
     get normalized() {
-        return this._valueToFilteringParamsMap.get(this._selectionDropdownModel.selected[0]) ?? null;
+        const [selectedOption] = this.selected;
+        return keyToMagnetsCurrentLevels(selectedOption);
     }
 
     /**
-     * Return the underlying selection dropdown model
+     * Sets selected options based on a comma-seperated string.
```
isaachilly: Expects { l3, dipole } object not comma-seperated string.

lib/public/components/Filters/RunsFilter/MagnetsFilteringModel.js
```diff
@@ -11,86 +11,57 @@
...
+        super(magnetsCurrentLevels$, magnetsCurrentLevelsToKey, { multiple: false });
     }
 
     /**
      * @inheritDoc
      */
     get normalized() {
-        return this._valueToFilteringParamsMap.get(this._selectionDropdownModel.selected[0]) ?? null;
+        const [selectedOption] = this.selected;
+        return keyToMagnetsCurrentLevels(selectedOption);
```
isaachilly, line 59: If nothing is selected will this throw an error?

lib/public/components/common/selection/SelectionModel.js
NarrowsProjects: Fair, I will place a listener on it's success if it is remoteData

lib/public/components/common/selection/SelectionModel.js
```diff
@@ -332,6 +351,23 @@ export class SelectionModel extends Observable {
         return this._defaultSelection;
     }
 
+    /**
+     * Sets selected options based on a comma-seperated string.
+     * Accounts for the options being either RemoteData or an array.
+     *
+     * @return {string}
```
NarrowsProjects: This should be a param, not a return

lib/public/components/common/selection/SelectionModel.js
NarrowsProjects: Fair, especially since a whitespace would be unrecognizable when warnings will be implemented

lib/public/components/common/selection/SelectionModel.js
NarrowsProjects: Since queryRouter omits any values that are undefined when computing its `params` field, the normalized values derived from `params` will always be usable. 

As a result, if a value is passed via URL (which is why the setters exist), neither value nor its equivalent can end up being undefined.

lib/public/components/common/selection/SelectionModel.js
NarrowsProjects: Your point still stands, but I realize that selectedOptions is a setter, and it already does that, so only the backlog clear needs to be removed

lib/public/components/Filters/RunsFilter/MagnetsFilteringModel.js
NarrowsProjects: Normalize is only ever called if isEmpty returns false.
By returning null or undefined (which the url parser excludes) we would be hiding the bug

lib/public/components/common/selection/SelectionModel.js
NarrowsProjects: To clarify, while undefined can't really reach the setters, empty strings can

lib/public/components/Filters/RunsFilter/MagnetsFilteringModel.js
```diff
@@ -11,86 +11,57 @@
...
     }
 
     /**
-     * Return the underlying selection dropdown model
+     * Sets selected options based on an object containing l3 and dipole fields.
+     * Accounts for the options being either RemoteData or an array.
      *
-     * @return {SelectionDropdownModel} the dropdown model
+     * @param {object}
```
graduta: Incomplete docs

lib/public/components/Filters/RunsFilter/MagnetsFilteringModel.js
```diff
@@ -11,86 +11,57 @@
...
-        return this._selectionDropdownModel.isEmpty;
+        super(magnetsCurrentLevels$, magnetsCurrentLevelsToKey, { multiple: false });
     }
 
     /**
      * @inheritDoc
      */
     get normalized() {
-        return this._valueToFilteringParamsMap.get(this._selectionDropdownModel.selected[0]) ?? null;
+        const [selectedOption] = this.selected;
```
graduta, line 53: Would it not be good to have a default value of empty in this case so that the mapping in `keyToMagnetsCurrentLevels` does not throw?

lib/public/components/common/selection/SelectionModel.js
```diff
@@ -332,6 +350,23 @@ export class SelectionModel extends Observable {
...
 
+    /**
+     * Sets selected options based on a comma-seperated string.
+     * Accounts for the options being either RemoteData or an array.
+     *
+     * @param {string} value the value that is to be set.
+     */
+    set normalized(value) {
+        const options = value.split(',').map((option) => ({ value: option.trim() }));
+        const postponeSelection = this.options instanceof RemoteData || ! this.options?.length;
```
graduta: Extra space after negation

lib/public/components/Filters/common/TagFilterModel.js
```diff
@@ -58,11 +58,19 @@ export class TagFilterModel extends FilterModel {
      */
     get normalized() {
         return {
-            values: this.selected.join(),
+            values: this._selectionModel.normalized,
             operation: this.combinationOperator,
```
graduta: In lib/public/components/Filters/common/filters/TextComparisonFilterModel.js you normalized the getter of operation as well. Why is that not the case here?

graduta (commented):
Nice effort, seems to be in the right direction. I wanted to ask if you had the opportunity to look at tests or update them? More specifically, given the refactor, what was your testing strategy to ensure consistency across changes?

lib/public/components/Filters/common/TagFilterModel.js
NarrowsProjects: There is no reason for that, I missed it

NarrowsProjects:
> Nice effort, seems to be in the right direction. I wanted to ask if you had the opportunity to look at tests or update them? More specifically, given the refactor, what was your testing strategy to ensure consistency across changes?

The refactors were all in models that already had tests that checked the effectiveness. 
The only thing of note that changed was how the runTypes filter had the return type of it's normalized function changed.

However, according to the the back-end code. And the test that checks the effectiveness of the runTypes filter on the front-end, this change did not make a difference.

So, I'm not quite sure which tests would be updated.

graduta (approved):

PR #2144, opened 2026-04-27
[O2B-1558] Set filters from url upon page reload

NarrowsProjects:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- When a filters on overview pages are set, the active filters will appear in the url, when reloading the page, these filters will be applied automatically.
- Users can now 'undo' filters by pressing the go-back button in the browser

- The eorReason filter component has been changed to match its reasonType and Category with its model
- A textInputFilter has been replaced by a textFilter as the model that controls it is a TextTokenFilterModel.
  - This prevented it from being set by the url as their state is different.
- A guard has been added in the registerObservablesQcSummaryDependsOn function to prevent continual re-registration
- A new getter has been added to filterModel and selectionModel called isInactive that will by default act as an alias for the isEmpty getter
  - The distinction is that isEmpty is used to determine if the filter should be added to requests or be set to the url while isInactive determines if specific ui interactions should happen (e.g. toggles being considered inactive despite being filled with 'false', thus being added urls, but not enabling the reset-filter button, as you can't reset an already untoggled toggle)
- The NumericalComparisonFilter normalized setter has been changed to separate raw value from parsed value.

- ToggleFilters will now ALWAYS be filled, meaning that they will always be added to the query parameters.
  - This was already the case for all toggles except for HasStableBeams, which would only be added if it was true.
  - The change does not change page behaviour as the back-end checks hasStableBeams for being truethy, and both false and undefined (which it was before this change) rule as falsely.

Changes made to the database:
- N/A


lib/public/components/Filters/common/FilteringModel.js
```diff
@@ -140,6 +140,30 @@ export class FilteringModel extends Observable {
         this.notify();
     }
 
+    /**
+     * Look for parameters used for filtering in URL and apply them in the layout if it exists
+     * @returns {undefined}
+     */
+    async setFilterFromURL() {
```
graduta: I do not think we need this to be `async`

lib/public/components/Filters/common/FilteringModel.js
```diff
@@ -140,6 +140,30 @@ export class FilteringModel extends Observable {
...
+        const { params: { page = '', filter = {} } } = this._router;
+
+        if (!(this._pageIdentifier === page)) {
```
graduta: I think it is better from readability's point of view that we use: `this._pageIdentifier !== page`

lib/public/components/Filters/common/FilteringModel.js
```diff
@@ -140,6 +140,30 @@ export class FilteringModel extends Observable {
...
+            return;
+        }
+
+        for (const [key, value] of Object.entries(filter)) {
+            const filterModel = this._filters[key];
```
graduta: What about this instead:
```js
            if (this._filters[key]) {
                this._filters[key].normalized = value;
            }
```

test/public/Filters/index.js
```diff
@@ -12,7 +12,9 @@
  */
 
 const ToUrlSuite = require('./filtersToUrl.test.js');
+const ToFilterSuite = require('./urlToFilter.test.js');
 
 module.exports = () => {
-    describe('Filters to URL', ToUrlSuite);
+    // describe('Filters to URL', ToUrlSuite);
```
graduta: Mistake in commenting out ?

lib/server/controllers/dataPasses.controller.js
```diff
@@ -25,11 +25,16 @@ const {
...
     NON_PHYSICS_PRODUCTIONS_NAMES_TOTAL_LENGTH,
 } = require('../../domain/enums/NonPhysicsProductionsNamesWords.js');
+const FilterLogger = require('../Loggers/FilterLogger.js');
 
 /**
  * List All DataPasses with statistics
  */
 const listDataPassesHandler = async (req, res) => {
+    const { path, session: { id }, query } = req;
+    const filters = query?.filters;
```
graduta: I believe it is `filter` without the `s`

lib/public/views/Runs/Overview/RunsWithQcModel.js
```diff
@@ -73,6 +73,10 @@ export class RunsWithQcModel extends RunsOverviewModel {
     constructor(model, pageIdentifier) {
         super(model, pageIdentifier);
 
+        this._detectorsNotBadFractionRegistered = false;
+        this._detectorsForQcFlagRegistered = false;
+        this._qcSummaryObservableRegistered = false;
```
graduta: This variable does not exist anywhere else in the code. What is its purpose?

lib/server/Loggers/FilterLogger.js
```diff
@@ -0,0 +1,51 @@
...
+        } else {
+            message += 'with the following filters:\n';
+            message += JSON.stringify(filters);
+        }
+
+        this._logger.infoMessage(message, { level: this._logLevel });
+    }
+}
+
+module.exports = new FilterLogger;
```
graduta: Missing bracket initialization

lib/public/views/Runs/RunsPerSimulationPass/RunsPerSimulationPassOverviewModel.js
```diff
@@ -75,21 +76,16 @@ export class RunsPerSimulationPassOverviewModel extends FixedPdpBeamTypeRunsOver
...
+        return buildUrl(super.getRootEndpoint(), { filter });
     }
 
     /**
      * Set id of current simulation pass which runs are fetched
      * @param {number} simulationPassId simulation pass id
      */
     set simulationPassId(simulationPassId) {
-        if (simulationPassId !== this._simulationPassId) {
+        if (this._simulationPassId && simulationPassId !== this._simulationPassId) {
```
graduta, line 87: Is this extra check needed?

lib/public/views/Runs/RunPerPeriod/RunsPerLhcPeriodOverviewModel.js
```diff
@@ -152,7 +148,7 @@ export class RunsPerLhcPeriodOverviewModel extends FixedPdpBeamTypeRunsOverviewM
      * @param {string} lhcPeriodId id of a LHC period
      */
     set lhcPeriodId(lhcPeriodId) {
-        if (lhcPeriodId !== this._lhcPeriodId) {
+        if (this._lhcPeriodId && lhcPeriodId !== this._lhcPeriodId) {
```
graduta, line 150: Is this extra check needed?

lib/public/views/Runs/RunPerDataPass/RunsPerDataPassOverviewModel.js
```diff
@@ -272,7 +273,7 @@ export class RunsPerDataPassOverviewModel extends FixedPdpBeamTypeRunsOverviewMo
      * @param {number} dataPassId id of Data Pass
      */
     set dataPassId(dataPassId) {
-        if (dataPassId !== this._dataPassId) {
+        if (this._dataPassId && dataPassId !== this._dataPassId) {
```
graduta, line 275: Is this extra check needed?

graduta (changes requested):
Comment for general description of the PR, it does more than what you mention:
- changes the use of textInput component
- changes EOR
- adds the logger

Comments for the logger - as you are introducing lots of duplicated code, I belive this would be better to be added as a middleware method and separate ticket
Comment for the call to set from URL. Could this not be added in a common space, perhaps something like in the setter of pageIdentifier?

lib/public/components/Filters/common/FilteringModel.js
NarrowsProjects: how about: 
```js
            if (key in this._filters) {
                this._filters[key].normalized = value;
            }
```

lib/public/views/Runs/Overview/RunsWithQcModel.js
NarrowsProjects: It serves no purpose anymore

lib/public/views/Runs/RunPerPeriod/RunsPerLhcPeriodOverviewModel.js
NarrowsProjects: Yes

The first time the page loads this setter is called.
The filters are set before `reset` is called.
This will immediately erase the active filters.

lib/public/views/Runs/RunsPerSimulationPass/RunsPerSimulationPassOverviewModel.js
NarrowsProjects: See response https://github.com/AliceO2Group/Bookkeeping/pull/2144/changes#r3310711170

lib/server/controllers/dataPasses.controller.js
NarrowsProjects: You're right

test/public/Filters/index.js
NarrowsProjects: indeed

lib/public/views/Runs/RunPerDataPass/RunsPerDataPassOverviewModel.js
NarrowsProjects: See response https://github.com/AliceO2Group/Bookkeeping/pull/2144/changes#r3310711170

lib/public/views/Environments/Overview/EnvironmentOverviewModel.js
```diff
@@ -92,7 +92,7 @@ export class EnvironmentOverviewModel extends OverviewPageModel {
     /**
      * Reset all filtering models
      * @param {boolean} fetch Whether to refetch all data after filters have been reset
-     * @param {boolean} [clearUrl=false] if true filters will be removed from the url
+     * @param {boolean} resetUrl Whether to remove all the active filters from the urls
      * @return {void}
      */
     resetFiltering(fetch = true, clearUrl = false) {
```
graduta, line 106: I think you updated the docs from the other file similar changes, but you did not apply the change to the implementaiton

graduta (changes requested):
I believe I identified a bug that needs to be taken care of:
- open a summary page (let's say logs)
- set a filter (let's say "No Anonymous")
- pages succesfully gets updated and the URL as well
- use the browser "Go Back" navigation button
- URL gets updated but the request towards the backend will still contain the filter applied

When fixing the test, please add a test for this scenario as well

graduta (approved):

PR #2145, opened 2026-04-28
[O2B-1559] Warn the user of mutated filters when loading the url

NarrowsProjects:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- When the user navigates to an overview page with filters that do not exist or cannot be parsed, the page will have an orange bar below the toolbar that says 'warnings' along with the relevant warnings

- **Caviat**:
  - The warnings have greater persistence than the faulty filters (which will disapear after configuring another filter)
  - The tickets: https://its.cern.ch/jira/browse/O2B-1581 and https://its.cern.ch/jira/browse/O2B-1593 address this issue

- A map is created inside OverviewModel, that is passed to filteringModel, where it will set messages about missing filters and unparsable filters

Changes made to the database:
- N/A

### DIFF
--- a/lib/public/components/Filters/common/FilteringModel.js
+++ b/lib/public/components/Filters/common/FilteringModel.js
@@ -153,7 +153,7 @@ export class FilteringModel extends Observable {
                 const filterModel = this._filters[key];
 
                 if (!filterModel) {
-                    unknownFilters.push(key);
+                    unknownFilters.push(`'${key}'`);
                     continue;
                 }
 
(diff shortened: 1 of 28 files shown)


lib/public/components/Filters/common/FilteringModel.js
```diff
@@ -142,22 +144,30 @@ export class FilteringModel extends Observable {
...
+                    unknownFilters.push(`'${key}'`);
                 }
-            } else {
-                this.reset();
             }
-        }
 
-        if (notify) {
-            this.notify();
+            if (unknownFilters.length) {
```
graduta: I would recommend to be explicit here and check length is positive.
I know that from a boolean logic, 0 is treated as false, but in this case a number comparison is more easy to read

graduta (changes requested):
Looks good, from a user experience point of view. Good job!

I have a few comments for you to think about:
- what do you think rather than adding a warning component on each page component, to instead define a parent `table` component that includes this warning and use this on the overview pages that need warnings?
- I discovered a small bug while testing. I will let you decide if it is worth fixing in this PR or a new ticket should be created. Basically, if a filter that expects a `values` sub-filter is not there, it will throw an error as to unable to `split` on undefined value. This error is not displayed in the GUI and instead is silently thrown in the console.

PR #2146, opened 2026-04-29
[O2B-1561] Alter filter popup button to have a drop down function where the user can copy, paste and clear the filters

NarrowsProjects:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- The filter toggle button how has a drop-down next to it that allows them to copy, paste, and reset filters

- FilteringModel now has an enum with the keys for the different warnings it handles
- A new warning type has been added where the user tries to paste a filter string on the wrong page
- FilteringModel.setFiltersFromURL has a new 'url' argument that if set, will be used in place of the router.params
- A significant refactor has been done on theSetFiltersFromURL functioni.
  - The filter setting loop has been moved to it's own function called _setFilters
  - The warnings are now cleared before anything else and re-applied as needed
- All OverviewPageModels have the setFiltersFromURL function now.

Changes made to the database:
- N/A

### DIFF
--- a/lib/public/components/Filters/common/FilteringModel.js
+++ b/lib/public/components/Filters/common/FilteringModel.js
@@ -168,7 +168,7 @@ export class FilteringModel extends Observable {
      * @param {string|null} [url=null] the url that is to be parsed into active filters
      * @returns {undefined}
      */
-    setFilterFromURL(url = null, notify = false) {
+    setFilterFromURL(notify = false, url = null) {
         const params = this._computeParameters(url);
         const { page, filter } = params;
 
--- a/lib/public/components/Filters/common/FilteringModel.js
+++ b/lib/public/components/Filters/common/FilteringModel.js
@@ -169,7 +169,8 @@ export class FilteringModel extends Observable {
      * @returns {undefined}
      */
     setFilterFromURL(url = null, notify = false) {
-        const { params: { page = '', filter } } = this._router;
+        const params = this._computeParameters(url);
+        const { page, filter } = params;
 
         if (this._pageIdentifier === page) {
             if (!filter) {
--- a/lib/public/components/Filters/common/FilteringModel.js
+++ b/lib/public/components/Filters/common/FilteringModel.js
@@ -37,7 +37,6 @@ export class FilteringModel extends Observable {
         this._filters = {};
         this._filterModels = [];
         Object.entries(filters).forEach(([key, model]) => this.put(key, model));
-        this.previousUrl = '';
     }
 
     /**
@@ -62,8 +61,6 @@ export class FilteringModel extends Observable {
         if (!this.isAnyFilterActive()) {
             return;
         }
-
-        this.previousUrl = location.href;
 
         for (const model of this._filterModels) {
             model.reset();
@@ -142,15 +139,6 @@ export class FilteringModel extends Observable {
         }
 
         this.notify();
-    }
-
-    /**
-     * Will restore filters to the state before it was reset.
-     * @returns {undefined}
-     */
-    restoreFilters() {
-        this.setFilterFromURL(this.previousUrl);
-        this.previousUrl = '';
     }
 
     /**
--- a/lib/public/components/Filters/common/FilteringModel.js
+++ b/lib/public/components/Filters/common/FilteringModel.js
@@ -16,6 +16,13 @@ import { SelectionModel } from '../../common/selection/SelectionModel.js';
 import { FilterModel } from './FilterModel.js';
 import { buildUrl, Observable, parseUrlParameters } from '/js/src/index.js';
 
+const WARNING_TYPES = Object.freeze({
+    PAGE_MISMATCH: 'Page-Filter mismatch',
+    UNKNOWN_FILTERS: 'Unknown Filters',
+    UNPARSABLE_URL: 'Unparseable URL',
+    UNPARSABLE_FILTERS: 'Unparsable Filters',
+})
+
 /**
  * Model representing a filtering system, including filter inputs visibility, filters values and so on
  */
@@ -71,6 +78,7 @@ export class FilteringModel extends Observable {
         }
 
         if (clearUrl) {
+            this._clearWarnings();
             const { params } = this._router;
             params.filter = this.normalized;
             this._router.go(buildUrl('?', params), false, true);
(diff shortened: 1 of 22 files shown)


graduta (changes requested):
While testing I discovered the following major issue:
1. Go to an overview table page, let's say runs
2. Add a filter option, any
3. Use the "Copy Active Filters" button from the filter dropdown
4. Navigate to another overview table page, let's say LhcFills
5. Use the "Paste" button from the filter dropdown.
6. Observe how the URL changed the parameter for the page to runs and filter parameters, but we are currently still on the page LHC Fills

NarrowsProjects:
> While testing I discovered the following major issue:
> 
> 1. Go to an overview table page, let's say runs
> 2. Add a filter option, any
> 3. Use the "Copy Active Filters" button from the filter dropdown
> 4. Navigate to another overview table page, let's say LhcFills
> 5. Use the "Paste" button from the filter dropdown.
> 6. Observe how the URL changed the parameter for the page to runs and filter parameters, but we are currently still on the page LHC Fills

I've changed the behavior to cancle the filter setting if it finds that the page of the url does not match up with the current page. 

In addition, I've added a warning for this specific scenario
image (1)

lib/public/components/Filters/common/FilteringModel.js
```diff
@@ -191,6 +214,43 @@ export class FilteringModel extends Observable {
         }
     }
 
+    /**
+     * Clear all filter-related warnings from the warnings map
+     *
+     * @returns {undefined}
+     */
+    _clearWarnings() {
+        this._warnings.delete(WARNING_TYPES.UNPARSABLE_URL);
```
graduta: As the WARNING_TYPES dictionary will potentially grow in size, why not make it a loop based delete for each key?

lib/public/components/Filters/common/filtersPanelPopover.js
```diff
@@ -114,13 +123,78 @@ const filtersToggleContent = (
...
+/**
+ * A button component that lets the user copy the url if there are active filters.
+ *
+ * @param {boolean} activeFilters if false, will disable the button
+ * @returns {Component} the copy button component
+ */
+const copyButtonOption = (activeFilters) => h(
+    '',
+    { style: activeFilters ? {} : { opacity: 0.5, pointerEvents: 'none' } },
+    h(CopyToClipboardComponent, { value: location.href, id: 'filter' }, 'Copy Active Filters'),
```
graduta: Is this ID value recommended 'filter'?

test/public/components/filtersPopoverPanel.test.js
```diff
@@ -0,0 +1,95 @@
...
+        await expectInputValue(page, '.name-filter input', '');
+        await expectInputValue(page, '.year-filter input', '');
+        await expectInputValue(page, '.pdpBeamTypes-filter input', '');
+    });
+
+    it('Should show warning if filters are pasted on the wrong page', async () => {
+        const url = 'http://localhost:4000/?page=lhc-period-overview&filter[names][]=name';
+        await goToPage(page, 'log-overview');
+
+        await page.evaluate(async (url) => await navigator.clipboard.writeText(url), url);
```
graduta: Why not simulating entirely the user steps and use the copy button instead?
I am recommending this for all above tests where it is similar

test/public/components/filtersPopoverPanel.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+        const actualUrl = page.url();
+        expect(actualUrl).to.equal(url);
+
+        await expectInputValue(page, 'div.flex-row.items-baseline:nth-of-type(1) input[type=text]', 'name');
+        await expectInputValue(page, 'div.flex-row.items-baseline:nth-of-type(2) input[type=text]', '100');
+        await expectInputValue(page, 'div.flex-row.items-baseline:nth-of-type(3) input[type=text]', 'PbPb');
+    });
+
+    it('Should set filters when pressing paste active filters button', async () => {
```
graduta: Is this test a duplicate of the one above?

test/public/components/filtersPopoverPanel.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+        const actualUrl = page.url();
+        expect(actualUrl).to.equal(url);
+
+        await expectInputValue(page, 'div.flex-row.items-baseline:nth-of-type(1) input[type=text]', 'name');
+        await expectInputValue(page, 'div.flex-row.items-baseline:nth-of-type(2) input[type=text]', '100');
+        await expectInputValue(page, 'div.flex-row.items-baseline:nth-of-type(3) input[type=text]', 'PbPb');
+    });
+
+    it('Should reset filters when pressing the reset all filters button', async () => {
```
graduta, line 53: I have seen you have a check for poorly copied filter values as well. Could you add a test for it? Or is it covered already by the tests of filtering?

graduta (changes requested):

lib/public/components/Filters/common/filtersPanelPopover.js
NarrowsProjects: I'm not sure what you mean?

I've changed it to be plural to match it with the reset buttons.
However, I'm not sure if that is what you meant.

The copyToClipBoardComponent looks at the context of the surrounding nodes and grabs the id value and parses it into `copy-${id}`. Meaning that the actual outputted ID is 'copy-filters'.

test/public/components/filtersPopoverPanel.test.js
NarrowsProjects: The test is more isolated this way. 
However, I do recognize that the rest of the front-end tests don't rigorously follow that principle.

So, I leave the decision up to you

test/public/components/filtersPopoverPanel.test.js
NarrowsProjects: Indeed it is

test/public/components/filtersPopoverPanel.test.js
NarrowsProjects: In addition to the test you suggested, I've also moved the test to the warning component suite, as I think in retrospect that both probably fit better there.

graduta (approved):

PR #2151, opened 2026-05-08
[O2B-1583] Sort lhc period overview page by id and not name

graduta:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- `unknown` datapass will not be displayed as first anymore

- page for `lhcPeriodOverview` was sorted by name because the naming convention implies also the chronological order of the periods. Because of this, 'u'>'l' was also at top.
- now, sorting is done by the period id, which as it is an incremental column, also shows chornological order of data pass

Changes made to the database:
- N/A

### DIFF
--- a/lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
+++ b/lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
@@ -85,11 +85,13 @@ class LhcPeriodStatisticsService {
         sort,
     } = {}) {
         const queryBuilder = this.prepareQueryBuilder();
-
         if (sort) {
             for (const property in sort) {
                 let expression;
                 switch (property) {
+                    case 'id':
+                        expression = (sequelize) => sequelize.col('`lhcPeriod`.`id`');;
+                        break;
                     case 'name':
                         expression = (sequelize) => sequelize.col('`lhcPeriod`.`name`');
                         break;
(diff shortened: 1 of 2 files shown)


lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
```diff
@@ -85,11 +85,13 @@ class LhcPeriodStatisticsService {
         sort,
     } = {}) {
         const queryBuilder = this.prepareQueryBuilder();
-
         if (sort) {
             for (const property in sort) {
                 let expression;
```
NarrowsProjects: Given that all the current expressions are hardcoded (and presumably all future expressions too), I'd argue that you could safely replace the switch statement with a lookup table where you pass the result to a sequelize.literal() function call.

```js

        const expressionMap = {
            id: '`lhcPeriod`.`id`',
            name: '`lhcPeriod`.`name`',
            year: 'SUBSTRING(lhcPeriod.name, 4, 2)',
            pdpBeamTypes: 'pdpBeamTypes',
        };

        if (sort) {
            for (const property in sort) {
                let expression;

                if (property in expressionMap) {
                    expression = (sequelize) => sequelize.literal(expressionMap[property]);
                }

                queryBuilder.orderBy(expression ?? property, sort[property]);
            }
        }
```

NarrowsProjects (changes requested):

lib/server/services/lhcPeriod/LhcPeriodStatisticsService.js
graduta: @NarrowsProjects  thank you for the suggestion!
I have updated the code but a bit different than what you suggesded because in your suggestion, you were losing the difference between `sequalize.literal` and `sequalize.col`

NarrowsProjects (approved):

PR #2160, opened 2026-06-01
[O2B-1592] Add infologger listeners middleware to overview page endpoints

NarrowsProjects:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- None

- A singleton called FilterLogger has been added which sends data from requests to InfoLogger.
- A middleware generator has been added which receives a type of logger that must implement the infoMessage function.
- So long as the node environment is not set to test, requests from the overview pages along with their filters, and the user who sent them are sent to InfoLogger.

Changes made to the database:
- None


lib/server/Loggers/FilterLogger.js
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Logs an informational message about endpoint access and applied filters.
+     *
+     * @param {object} request the request received at any given endpoint.
+     * @param {string} endpoint the endpoint that was accessed.
+     * @param {string|number} id identifier of the user accessing the endpoint.
+     * @param {Object} [filters={}] filters applied to the request.
+     * @returns {void}
+     */
+    infoMessage({ path, session: { id }, query = {} }) {
```
graduta: I believe you need to ensure default values for the session otherwise the method will throw on deconstructing the JSON

lib/server/Loggers/FilterLogger.js
```diff
@@ -0,0 +1,59 @@
...
+        if (this._silent) {
+            return;
+        }
+
+        const filters = query.filters ?? {};
```
graduta: Is it `filters` or `filter`? Please have a look at front-end code

graduta (changes requested):

lib/server/Loggers/FilterLogger.js
NarrowsProjects: Darn, same mistake I made last time.

lib/server/Loggers/FilterLogger.js
NarrowsProjects: Fair.

The usecase where there is no session is when an unauthenticated user accesses the endpoint. 
So, in addition to the change you suggested, I'll add another message construction path that takes note of that scenario when it happens.

graduta (approved):

PR #2172, opened 2026-06-19
[O2B-1595] Add filterable overview page class

NarrowsProjects:
#### I DON'T have JIRA ticket
- [ ] tests are added

A new class called FilterableOverviewPageModel has been created. 
All classes that previously instantiated a FilteringModel in them have now been changed to extend this class instead.

Notable changes for users:
- N/A

- The FilterableOverviewPageModel class by default fetches it's data immediately. 
- If you wish to fetch it using debounce, you'll have to set the variable 'fetchImediately' to true
- The debounce functionallity from all overview pages has been removed
  - The reason is that the pages define the functionallity, but don't use it*
  - \* LogsOverview does use it, but exclusively for sorting, for which the effective change is very minimal.

Changes made to the database:
- N/A

### DIFF
--- a/lib/public/models/FilterableOverviewPageModel.js
+++ b/lib/public/models/FilterableOverviewPageModel.js
@@ -23,6 +23,9 @@ import { FilteringModel } from '../components/Filters/common/FilteringModel.js';
 export class FilterableOverviewPageModel extends OverviewPageModel {
     /**
      * Constructor
+     * @param {QueryRouter} router router that controls the application's page navigation
+     * @param {string} pageIdentifier string that indicates what page this model represents
+     * @param {Object<string, FilterModel>} filters the filters with their label and model
      */
     constructor(router, pageIdentifier, filters) {
         super();
--- a/lib/public/models/FilterableOverviewPageModel.js
+++ b/lib/public/models/FilterableOverviewPageModel.js
@@ -34,6 +34,7 @@ export class FilterableOverviewPageModel extends OverviewPageModel {
         this._filteringModel.pageIdentifier = pageIdentifier;
         this._filteringModel.visualChange$.bubbleTo(this);
         this._filteringModel.observe(() => this._applyFilters());
+        this._sortModel.unobserve(this._sortModelCallback);
         this._sortModel.observe(() => this._applyFilters());
         this._debouncedLoad = (_time) => {}; // Abstract, does nothing on purpose
         this._fetchInstantly = true;
--- a/lib/public/models/FilterableOverviewPageModel.js
+++ b/lib/public/models/FilterableOverviewPageModel.js
@@ -1,0 +1,128 @@
+/**
+ * @license
+ * Copyright CERN and copyright holders of ALICE O2. This software is
+ * distributed under the terms of the GNU General Public License v3 (GPL
+ * Version 3), copied verbatim in the file "COPYING".
+ *
+ * See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+import { buildUrl } from '/js/src/index.js';
+import { OverviewPageModel } from './OverviewModel.js';
+import { FilteringModel } from '../components/Filters/common/FilteringModel.js';
+
+/**
+ * Base model for a filterable overview page
+ *
+ * @template T the type of data displayed in the overview page
+ */
+export class FilterableOverviewPageModel extends OverviewPageModel {
+    /**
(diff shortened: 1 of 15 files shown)


lib/public/models/FilterableOverviewPageModel.js
graduta, line 1: The whole file introduces some sort of eslint warnings due to incomplete documentation

lib/public/views/Runs/Overview/RunsOverviewModel.js
```diff
@@ -122,81 +114,27 @@ export class RunsOverviewModel extends OverviewPageModel {
...
-    /**
-     * Returns all filtering, sorting and pagination settings to their default values
-     * @param {boolean} [fetch = true] whether to refetch all data after filters have been reset
-     * @return {void}
-     */
     reset(fetch = true) {
         super.reset();
         this._exportModel?.reset();
         this.resetFiltering(fetch);
     }
```
graduta, line 115: I believe that `super.reset()` in this case will also already call `this.resetFiltering`
Of course the difference is the value of `fetch` but still seems like a duplication

lib/public/views/lhcPeriods/Overview/LhcPeriodsOverviewModel.js
```diff
@@ -11,33 +11,29 @@
...
+            pageIdentifier,
             {
                 names: new TextTokensFilterModel(),
                 years: new TextTokensFilterModel(),
                 pdpBeamTypes: new TextTokensFilterModel(),
             },
-            this._warnings,
         );
 
         this._filteringModel.pageIdentifier = pageIdentifier;
```
graduta: Give that the filtering model is now on its own, why do we have to still do the pageIndentifier and rest of commands related here?

lib/public/views/Logs/Overview/LogsOverviewModel.js
```diff
@@ -57,75 +54,13 @@ export class LogsOverviewModel extends OverviewPageModel {
 
         updateDebounceTime();
```
graduta: The debounce functionality does not work anymore. Was this on purpose?
I agree that all fitlers should behave similarly (so at out of focus) and we can move the log filtering to that usecase as well

lib/public/models/FilterableOverviewPageModel.js
```diff
@@ -0,0 +1,128 @@
...
+     * Constructor
+     */
+    constructor(router, pageIdentifier, filters) {
+        super();
+        this._filteringModel = new FilteringModel(router, filters, this._warnings);
+
+        this._filteringModel.pageIdentifier = pageIdentifier;
+        this._filteringModel.visualChange$.bubbleTo(this);
+        this._filteringModel.observe(() => this._applyFilters());
+        this._sortModel.observe(() => this._applyFilters());
```
graduta, line 38: We also observe the sortmodel in the OverviewModel, which leads to double loading (easy to test on logs page), right?

graduta (changes requested):

lib/public/models/FilterableOverviewPageModel.js
NarrowsProjects: I see that I missed the arguments from the constructor, but oddly, it is not caught by eslint on github, nor locally for me

lib/public/views/lhcPeriods/Overview/LhcPeriodsOverviewModel.js
NarrowsProjects: good catch, I missed that one

lib/public/views/Logs/Overview/LogsOverviewModel.js
NarrowsProjects: It was not on purpose, but the main use case for why LogsOverview needed the debounce was moved away from in an earlier branch.

So, very little is lost in removing it entirely from the LogsOverview model

lib/public/models/FilterableOverviewPageModel.js
NarrowsProjects: Yeah, I missed that, however, strangely, whilst the load() function runs twice, only one request seems to be sent out.

lib/public/views/Logs/Overview/LogsOverviewModel.js
NarrowsProjects: Update: I've gone over the other pages again, and whilst most of them do define debounce functionality, none of them actually use it anywhere

lib/public/models/FilterableOverviewPageModel.js
graduta: How are you testing it? Are you sure you are triggering the sort model? That one is per column header click in the table only. So in Logs for example but not in other tables

graduta (approved):

PR #2200, opened 2026-07-29
[O2B-1601] Add feature to allow users to filter runs by beam type

graduta:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- on run-overview page, users are now able to select from the widget boxes and filter by the beam type associated to the fill of that run.

### DIFF
--- a/lib/domain/dtos/common/BeamTypeDto.js
+++ b/lib/domain/dtos/common/BeamTypeDto.js
@@ -1,0 +1,23 @@
+/**
+ *  @license
+ *  Copyright CERN and copyright holders of ALICE O2. This software is
+ *  distributed under the terms of the GNU General Public License v3 (GPL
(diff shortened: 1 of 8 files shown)


isaachilly (commented):
The filter is called `PDP Beam Type` but its options and related filtering in the useCase is all related to `beamType`, as is the PR description but not the ticket description. I don't know which way around it's meant to be but I think right now they are at loggerheads?

graduta:
> The filter is called `PDP Beam Type` but its options and related filtering in the useCase is all related to `beamType`, as is the PR description but not the ticket description. I don't know which way around it's meant to be but I think right now they are at loggerheads?

You are absolutely right! The PDP Beam Type, which is an environment configurable parameter is next filter I will add. While the values correspond to the same idea, they contain different values. Thank you for noticing it!

isaachilly (approved):

PR #2202, opened 2026-07-31
[O2B-1602] Shared BeamTypesDto should return split input rather than simply validation

graduta:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- N/A

- BeamTypesFilterDto will now as per the convention return an array of beam types rather than a string which then every usecase (GetAllLhcFills and GetAllRuns) have to split the input
- tests adapted accordingly


lib/domain/dtos/common/BeamTypeDto.js
```diff
@@ -12,12 +12,17 @@
  */
 
 const Joi = require('joi');
-const { validateBeamTypes, BEAM_TYPE_INVALID } = require('../../../utilities/beamTypeUtils');
+const { CustomJoi } = require('../CustomJoi.js');
```
isaachilly, line 15: I understand this was not introduced in this PR, so **not** to be implemented here, but would you agree that `CustomJoi` is a rather nondescript name and should be changed to something like `StringArrayJoi? If so, I will make a note and then the change.

lib/domain/dtos/common/BeamTypeDto.js
```diff
@@ -12,12 +12,17 @@
...
-    .messages({
-        [BEAM_TYPE_INVALID]: '{{#message}}',
-        'string.base': 'Beam type must be a string',
-    });
+const BEAM_TYPE_INVALID = 'beamType.invalid';
+
+exports.BeamTypesDto = CustomJoi.stringArray()
+    .items(Joi.string()
+        .trim()
+        .min(2)
```
isaachilly: Is this minimum check necessary as the pattern enforces >2 chars with `char dash char`?

lib/domain/dtos/common/BeamTypeDto.js
```diff
@@ -12,12 +12,17 @@
...
+        .max(15)
+        .pattern(/^[A-Za-z0-9]+ ?- ?[A-Za-z0-9]+$/)
+        .messages({
```
isaachilly, line 40: The custom validator was the only thing raising `BEAM_TYPE_INVALID` error so I believe this is now dead message handler code.

Supported by the fact the `test/api/runs.test.js` LN196 test has been changed to assert the default JOI message format.

Do you agree?

isaachilly (changes requested):

lib/domain/dtos/common/BeamTypeDto.js
graduta: I do not believe it should be changed to "StringArrayJoi" because the name CustomJoi refers to the wrapper defined in that file.
If you look at the implementation of the wrapper, you will see it defines the method `stringArray` which is being called in this file. 
Thus, to me:
- "CustomJoi" - informs the developer that it is a custom wrapper around the Joi library
- "CustomJoi.stringArray()" - it is the call that returns the string array

lib/domain/dtos/common/BeamTypeDto.js
graduta: While the pattern will enforce that, I think it is also important to be self-explanatory when writing code, so that it is easy to maintain it.
Moreover, regexes are in general slow, so in this case a length-based match may be faster for scenarios in which we receive thousands of requests within a second.

lib/domain/dtos/common/BeamTypeDto.js
graduta: You are correct. I have modified the DTO to have more specific error messages now

lib/domain/dtos/common/BeamTypeDto.js
isaachilly: Ah okay, it is a normal Joi instance extended with the `stringArray` type, it just currently happens to only define one custom type. Makes sense, I withdraw the rename suggestion.

lib/domain/dtos/common/BeamTypeDto.js
isaachilly: Performance-wise we are agreed, it would be more efficient at rejecting than the regex however our `dtoValidator` has `abortEarly` set to `false` so it still tries every case in items. As evidenced by the API response to a one character beam type.

If kept, should it also be `min(3)` rather than 2, since `a-b` is the shortest the regex can match?

---

Response when GET-ing `http://localhost:4000/api/runs?page%5Boffset%5D=0&page%5Blimit%5D=18&filter%5BbeamTypes%5D=p&token=TOKENHERE`

```{
  "errors": [
    {
      "status": "422",
      "source": {
        "pointer": "/data/attributes/query/filter/beamTypes/0"
      },
      "title": "Invalid Attribute",
      "detail": "Beam type must be at least 2 characters long"
    },
    {
      "status": "422",
      "source": {
        "pointer": "/data/attributes/query/filter/beamTypes/0"
      },
      "title": "Invalid Attribute",
      "detail": "Beam type must look like \"PROTON - PROTON\", \"NE10 - NE10\", etc."
    }
  ]
}
```

graduta: Indeed, I missed the abortEarly from the `dtoValidator.js`
I cannot think of a reason why we should not abortEarly as of now so will take note of it. 

However, I prefer to keep the min/max for now as it is easier to check and also will be beneficial when we move to abortEarly to `true`

isaachilly: Agreed, I'm not suggesting we drop it. My point was at it stands, the `min`, and for that matter the`max`, aren't forcing an exit with `abortEarly` set the way it is, so we are not gaining the performance benefit... _yet_ and we should look into this.

Just the JSDoc and error message which need to be updated from 2 to 3 characters also and I'll approve 😊.

isaachilly (approved):

PR #2203, opened 2026-07-31
[O2B-1602] Add endpoint retrieval of pdp beam types and runs endpoint filtering option

graduta:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- N/A

- add new runs/pdpBeamType endpoint to fetch all uniquely declared pdpBeamTypes when deploying an environment/starting a run
- add option for GET runs endpoint to filter by pdpBeamType


test/lib/server/services/pdpBeamTypes/index.js
```diff
@@ -0,0 +1,18 @@
...
+ * See http://alice-o2.web.cern.ch/license for full licensing information.
+ *
+ * In applying this license CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+const getAllPdpBeamTypes = require('./getAllPdpBeamTypes.test.js');
+
+module.exports = () => {
```
isaachilly, line 16: Should this test suite be added to the `services/index.js` file?

test/lib/server/services/pdpBeamTypes/getAllPdpBeamTypes.test.js
```diff
@@ -0,0 +1,30 @@
...
+    it('should successfully return the full list of not null PDP beam types from runs table', async () => {
+        const pdpBeamTypes = await getAllPdpBeamTypes();
+        expect(pdpBeamTypes.map(({ dataValues: { name } }) => ({ name }))).to.deep.eq([
+            { name: 'cosmic' },
+            { name: 'technical' },
+            { name: 'pp' },
+            { name: 'PbPb' },
+            { name: 'pO' },
+            { name: 'OO' },
+            { name: 'NeNe' },
```
isaachilly: I ran this test and I think it fails because `pO` and `NeNe` are not in the seeded test database. Do you agree?

lib/domain/dtos/common/BeamTypeDto.js
```diff
@@ -14,10 +14,25 @@
...
+const BEAM_TYPE_MIN_LENGTH = 3;
+const BEAM_TYPE_MAX_LENGTH = 15;
+const PDP_BEAM_TYPE_MIN_LENGTH = 2;
+const PDP_BEAM_TYPE_MAX_LENGTH = 10;
+
 /**
  * @typedef {string[]} BeamTypesDto
  * @description An array of beam types, each represented as a string.
- * Each beam type must be a string with a minimum length of 3 characters and a maximum length of 15 characters.
+ * Each beam type must be a string with a minimum length of 2 characters and a maximum length of 15 characters.
```
isaachilly: The JSDocs and actual constraints are a little confusing to follow. What do you think?

- `BeamTypesDto` JSDoc now says 2 characters but the constant is still 3.
- `BeamTypesDto` says RUN3 has compact beam types but we validate against a pattern that would reject them.
- `PdpBeamTypesDto` has 10 as the max length constant but the regex independently allows 15. What do you think about adding the length constants directly into the regex so there's one source of truth?

test/api/runs.test.js
```diff
@@ -196,6 +196,52 @@ module.exports = () => {
...
+
+            expect(response.status).to.equal(200);
+            const { data: runs } = response.body;
+
+            expect(runs).to.be.an('array');
+            expect(runs).to.have.lengthOf(10);
+            expect(runs.every(({ pdpBeamType: type }) => pdpBeamTypes.includes(type))).to.be.true;
+        });
+
+        it('should return 400 if pdpBeamTypes filter has the incorrect format', async () => {
```
isaachilly, line 223: For completeness, what do you think about adding the test case of it's got a digit in it?

isaachilly (changes requested):

lib/domain/dtos/common/BeamTypeDto.js
graduta: Indeed. My thoughts initially was to have a generalized condition for both scenarios but I realized that it potentially creates confusion for developers.

test/api/runs.test.js
graduta: Sure thing, added one

isaachilly (approved):

PR #2205, opened 2026-07-31
[O2B-1602] Add widget for runs-overview to filter by pdpBeamType

graduta:
#### I have a JIRA ticket
- [x] PR labels are selected

Notable changes for users:
- runs-overview page now gives option to the user to filter by pdp beam type from a set of pre-selected options
- the beamtype filter model expects now a provider (being beam types or PDP beam types)
- the beamtype filter model now extends the observable pattern as it loads remote data to fill in the filter


lib/public/components/Filters/LhcFillsFilter/BeamTypeFilterModel.js
```diff
@@ -11,26 +11,18 @@
...
-        beamTypesProvider.items$.observe(() => {
-            beamTypesProvider.items$.getCurrent().apply({
-                Success: (types) => {
-                    const beamTypes = types.map((type) => ({ value: type.beam_type }));
-                    this.setAvailableOptions(beamTypes);
-                },
-            });
-        });
+    constructor(beamTypes$) {
+        super(beamTypes$, ({ name, beam_type }) => ({ value: name ?? beam_type }));
```
isaachilly: Would it be worth changing the `getAllBeamTypes` service so the model can be cleaner and just deal with `name`?

isaachilly (commented):
It is out-of-scope for this ticket, but the `beamType` selection could be applied to the RCT `lhc-period-overview` page also in the future. Do you agree?

lib/public/components/Filters/LhcFillsFilter/BeamTypeFilterModel.js
graduta: Not within this ticket, no. We have to keep in mind that there are are clients of the API besides the front-end. Thus, we should change it with announcements and deprecation notification

isaachilly: I did think of sole API users, despite not writing it in the comment. To go over with you about how we change an API like this would be useful. I guess there is the other way around of course, changing `getAllPdpBeamTypes` to use `beam_type`.

graduta:
> It is out-of-scope for this ticket, but the `beamType` selection could be applied to the RCT `lhc-period-overview` page also in the future. Do you agree?

I agree but I have the remark that it is not `beamType` that RCT users are seeking. For them we display and they use `pdpBeamType`

isaachilly:
> > It is out-of-scope for this ticket, but the `beamType` selection could be applied to the RCT `lhc-period-overview` page also in the future. Do you agree?
> 
> I agree but I have the remark that it is not `beamType` that RCT users are seeking. For them we display and they use `pdpBeamType`

Yes, I'm saying `beamType` here as in general category, like `beamTypeFilterModel`.

isaachilly (dismissed):

lib/public/components/Filters/LhcFillsFilter/BeamTypeFilterModel.js
graduta: Okay, let's stick with `beam_type` and think of refactor in the next ticket

isaachilly (approved):

